PR #135, opened 2022-11-05
Artifact book Shuffle when item is removed

TracyTheBoy:
Hi,

I've been playing the map for a while and since this repo is public I thought I'd try solving a couple open issues.

This attempts to solve issue #121 

All Artifact pages are redrawn whenever an artifact is removed. This ensure that there are no empty pages, no half full pages and no gaps within a page. So it should cover all cases. 

There aren't that many artifacts from what i've seen so even though this is probably not ideal performance wise I figured it wouldn't be that costly anyway.

### DIFF
--- a/src/MacroTools/Frames/Books/ArtifactSystem/ArtifactBook.cs
+++ b/src/MacroTools/Frames/Books/ArtifactSystem/ArtifactBook.cs
@@ -25,9 +25,7 @@ namespace MacroTools.Frames.Books.ArtifactSystem
       height, bottomButtonXOffset, bottomButtonYOffset)
     {
       ArtifactManager.ArtifactRegistered += ArtifactCreated;
-      var firstPage = AddPage();
-      firstPage.Visible = true;
-      AddAllArtifacts();
+      AddPagesAndArtifacts();
       BookTitle = "Artifacts";
       LauncherParent = BlzGetFrameByName("UpperButtonBarQuestsButton", 0);
       Position = new Point(0.4f, 0.38f);
@@ -41,7 +39,6 @@ namespace MacroTools.Frames.Books.ArtifactSystem
         AddPage();
         lastPage = Pages.Last();
       }
-
       lastPage.AddArtifact(artifact);
       _pagesByArtifact.Add(artifact, lastPage);
       artifact.Disposed += OnArtifactDisposed;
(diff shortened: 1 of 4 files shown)


YakaryBovine:
This is awesome, thankyou. I think the approach you've taken is fine since you're right; this is not a performance-critical area of code.

Your solution well when there is only one Artifact page but it breaks down when there are more.

Before destruction:
image
image

After destroying Kelen's Dagger of Escape:
image
image
So the wrong items are showing and the page number looks like it's showing both 1 and 2.

src/MacroTools/Frames/Books/ArtifactSystem/ArtifactBook.cs
```diff
@@ -51,12 +49,31 @@ private void OnArtifactDisposed(object? sender, Artifact artifact)
...
+      FillGaps();
+    }
+
+    private void FillGaps()
+    {
+        foreach (var artifact in _pagesByArtifact.Keys)
+        {
+            artifact.Disposed -= OnArtifactDisposed;
+        }
+        Pages.Clear();
```
YakaryBovine, line 61:
```cs
  foreach (var page in Pages) 
    page.Dispose();
```
You're clearing the frames out of memory but you're not disposing of them, so the pages don't get deleted until the garbage collector arrives. As far as I can tell this fixes the entire problem.

src/MacroTools/Frames/Books/ArtifactSystem/ArtifactBook.cs
```diff
@@ -51,12 +49,31 @@ private void OnArtifactDisposed(object? sender, Artifact artifact)
     {
       _pagesByArtifact[artifact].RemoveArtifact(artifact);
       _pagesByArtifact.Remove(artifact);
+      FillGaps();
+    }
+
+    private void FillGaps()
```
YakaryBovine, line 55: This isn't just filling gaps, it's re-rendering the whole thing. Maybe we can just call it ReRender?

YakaryBovine (changes requested):
Looks good, just a couple of fixes to be made.

src/MacroTools/Frames/Books/ArtifactSystem/ArtifactBook.cs
TracyTheBoy: Good idea. I wasn't happy with the Name either but ReRender sounds good

src/MacroTools/Frames/Books/ArtifactSystem/ArtifactBook.cs
TracyTheBoy: Yes I figured that out too.
I am now disposing of the page and artifact frames properly I think.

YakaryBovine (approved):

YakaryBovine:
Well done, thanks. Feel free to message me on Discord if you need anything.

Issue #72, opened 2022-09-30
HelmOfDominationDropsWhenScourgeLeaves


YakaryBovine:
Recreate the following vJASS trigger in C#.
When Scourge leaves the game, the Helm of Domination should drop to the ground instantly.
https://github.com/AzerothWarsLR/AzerothWarsLR/blob/master/jass/Mechanics/Scourge/HelmOfDominationDropsWhenScourgeLeaves.J

PR #137, opened 2022-11-06
attempt to solve issue #72

TracyTheBoy:
Dropping all LK items, then moving the item a more easily accessible spot.

I would have preferred to use `DropItem(item)` but I couldn't get that to work.
Since LK only has one item I hope it is okay to do it this way.

I am also unsure about the way I wired up the event. Haven't seen it done this way in any of the other Mechanics.  All of them use triggers but this worked in my tests.

### DIFF
--- a/src/WarcraftLegacies.Source/Mechanics/Scourge/HelmOfDominationDropsWhenScourgeLeaves.cs
+++ b/src/WarcraftLegacies.Source/Mechanics/Scourge/HelmOfDominationDropsWhenScourgeLeaves.cs
@@ -1,22 +1,31 @@
-// using WarcraftLegacies.Source.Legends;
-// using WarcraftLegacies.Source.Setup;
-// using WarcraftLegacies.Source.Setup.FactionSetup;
-//
-// using static War3Api.Common;  using static MacroTools.GeneralHelpers;
-// {
-//   public class HelmOfDominationDropsWhenScourgeLeaves{
-//
-//     private static void Actions( ){
-//       if (GetTriggerFaction() == ScourgeSetup.FactionScourge && ArtifactSetup.ArtifactHelmofdomination.OwningUnit == LegendScourge.LegendLichking.Unit){
-//         SetItemPosition(ArtifactSetup.ArtifactHelmofdomination.Item, GetRectCenterX(gg_rct_LichKing), GetRectCenterY(gg_rct_LichKing));
-//       }
-//     }
-//
-//     public static void Setup( ){
-//       trigger trig = CreateTrigger();
-//       OnFactionGameLeave.register(trig);
-//       TriggerAddAction(trig,  Actions);
-//     }
-//
-//   }
-// }
+using WarcraftLegacies.Source.Setup.Legends;
+using WarcraftLegacies.Source.Setup;
+using WarcraftLegacies.Source.Setup.FactionSetup;
+using MacroTools.FactionSystem;
+using MacroTools.Extensions;
+using WCSharp.Shared.Data;
+
+/// <summary>
+///  When Scourge leaves the game, drop the Helm of Domination to an easily accessible position since the default position is hard to see.
+/// </summary>
+public class HelmOfDominationDropsWhenScourgeLeaves
+{
+    public static void Setup()
+    {
+        Faction.GameLeave += Faction_GameLeave;
+    }
+
+    private static void Faction_GameLeave(object? sender, Faction leavingFaction)
+    {
+        if (ArtifactSetup.ArtifactHelmofdomination == null || LegendScourge.LegendLichking?.Unit == null)
+            return;
+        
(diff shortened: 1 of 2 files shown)


TracyTheBoy:
This is what it looks like when the item is on the ground
image

YakaryBovine:
Awesome stuff. Using that event is completely correct; the only reason the other Mechanics use triggers is because they're hooking directly into WC3 API logic, whereas this is dependent on our own custom events.

src/WarcraftLegacies.Source/Mechanics/Scourge/HelmOfDominationDropsWhenScourgeLeaves.cs
```diff
@@ -1,22 +1,25 @@
...
+using WarcraftLegacies.Source.Setup.Legends;
+using WarcraftLegacies.Source.Setup;
+using WarcraftLegacies.Source.Setup.FactionSetup;
+using static War3Api.Common;
+using MacroTools.FactionSystem;
+using MacroTools.Extensions;
+
+public class HelmOfDominationDropsWhenScourgeLeaves
+{
+    public static void Setup()
```
YakaryBovine, line 10: Could we add a quick documentation document here? E.g.:
```cs
/// <summary>
///  When Scourge leaves the game, drop the Helm of Domination to an easily accessible position since the default position is hard to see.
/// </summary>
```

src/WarcraftLegacies.Source/Mechanics/Scourge/HelmOfDominationDropsWhenScourgeLeaves.cs
```diff
@@ -1,22 +1,25 @@
...
+    {
+        Faction.GameLeave += Faction_GameLeave;
+    }
+
+    private static void Faction_GameLeave(object? sender, Faction e)
```
YakaryBovine, line 15: I don't know if it's standard for C# events, but lets' call the Faction parameter something descriptive like `leavingFaction`.

src/WarcraftLegacies.Source/Mechanics/Scourge/HelmOfDominationDropsWhenScourgeLeaves.cs
```diff
@@ -1,22 +1,25 @@
...
+    {
+        if (e == ScourgeSetup.Scourge && ArtifactSetup.ArtifactHelmofdomination.OwningUnit == LegendScourge.LegendLichking.Unit)
+        {
+            var lkPos = LegendScourge.LegendLichking.Unit.GetPosition();
```
YakaryBovine, line 19: I think we should use fully expanded variable names - so `lichkingPosition` in this case.

src/WarcraftLegacies.Source/Setup/FactionSetup/ScourgeSetup.cs
```diff
@@ -120,7 +120,10 @@ public static void Setup()
       Scourge.AddPower(visionPower);
       
       Scourge.AddGoldMine(PreplacedUnitSystem.GetUnit(FourCC("ngol"), new Point(-4939, 18803)));
-      
+
+      // Mechanics
```
YakaryBovine, line 124: Let's see if we can get away with removing this comment. In my experience they're not maintained properly and I think what you've written is clear without it.

src/WarcraftLegacies.Source/Mechanics/Scourge/HelmOfDominationDropsWhenScourgeLeaves.cs
```diff
@@ -1,22 +1,25 @@
...
+            LegendScourge.LegendLichking.Unit.DropAllItems();
+            SetItemPosition(ArtifactSetup.ArtifactHelmofdomination.Item, lkPos.X - 55, lkPos.Y + 30); // Move item in front of the LK
```
YakaryBovine, line 21:
```suggestion
            ArtifactSetup.ArtifactHelmofdomination?.Item.SetPosition(new Point(lichKingPosition.X - 55, lichKingPosition.Y + 30));
```
Something like this will stop the compiler complaining about the possible null reference, and it's a little more standard (use Points where possible). But it is of course a little uglier.

YakaryBovine (changes requested):
The trigger works perfectly, just a few style nitpicks to maintain consistency with the codebase.

src/WarcraftLegacies.Source/Mechanics/Scourge/HelmOfDominationDropsWhenScourgeLeaves.cs
TracyTheBoy: Microsoft usually doesn't name the parameter anything but e. I think they expect people to look at the type directly. 
But I will follow your preferred style of course.

YakaryBovine (approved):
Great stuff.

Issue #62, opened 2022-09-30
QuestScepterOfTheQueen


YakaryBovine:
Recreate the following vJASS quest in C#. Note that this is actually 2 quests in one file. For the C# version we probably want 2 files.
https://github.com/AzerothWarsLR/AzerothWarsLR/blob/master/jass/Quests/QuestScepterOfTheQueen.j

TracyTheBoy:
Hi, I have a question regarding this.
I couldn't find the c# equivalent of `RescueNeutralUnitsInRect()` that is used in the vJASSversion.
Could you point me in the right direction on what to use here?

YakaryBovine:
Yep. I haven't made a helper method for that one because it's a lot easier to do in C#. So I've been doing something like:
```cs
                        var unitsInRect = new GroupWrapper().EnumUnitsInRect(RECT).EmptyToList();
			foreach (var unit in unitsInRect)
			{
				unit.Rescue(PLAYER);
			}
```

That said, we should move away from this pattern. There's a problem with it; what happens if we happen to move another neutral unit into that area over the course of the game? We'll rescue it when we didn't intend to. 
Check out the pattern here: https://github.com/AzerothWarsLR/WarcraftLegacies/blob/main/src/WarcraftLegacies.Source/Quests/Ironforge/QuestDarkIron.cs
It puts all the units into a group when the Quest is made, then Rescues only the units in that group when the Quest is completed or failed. It also hides the non-building units to save users some CPU cycles.

TracyTheBoy:
Thanks I now implemented it this way.
I am not sure on how to test that part of the quest though because there are no units in that area and I don't know how to place any for testing purposes.

YakaryBovine:
There aren't? That's odd, there should be a bunch of archers... I'll check it out. Your changes look correct though so feel free to make the PR.

PR #139, opened 2022-11-06
Add Scepter of the Queen related quests to Warsong and Sentinels  #62

TracyTheBoy:
As already discussed in the comments of #62 I had no way of testing whether the units changing owner upon completion of the quest is actually working.

I made a rectExtension class to prevent duplication of code used in both Scepter Quests. 
I feel like extension methods do not really fit here but I couldn't find any other helper classes that are used in the same way so I figured I stick to the pattern that's already there.

### DIFF
--- a/src/MacroTools/Extensions/RectExtensions.cs
+++ b/src/MacroTools/Extensions/RectExtensions.cs
@@ -1,0 +1,58 @@
+﻿using MacroTools.Wrappers;
+using static War3Api.Common;
+
+namespace MacroTools.Extensions
+{
+    /// <summary>
+    /// Provides a helpful set of extension methods for <see cref="rect"/>s.
+    /// </summary>
+    public static class RectExtensions
+    {
+        /// <summary>
+        /// Renders all units inside the specified area invulnerable/>
+        /// </summary>
+        /// <param name="area"></param>
+        public static void MakeUnitsInvulnerable(this rect area)
+        {
+            var shenGroup = CreateGroup();
+            GroupEnumUnitsInRect(shenGroup, area, null);
+            bool execute = true;
+            while (execute)
+            {
+                var unit = FirstOfGroup(shenGroup);
+                if (unit == null)
+                {
+                    execute = false;
+                    continue;
+                }
+                if (GetOwningPlayer(unit) == Player(GetPlayerNeutralPassive()))
+                {
+                    SetUnitInvulnerable(unit, true);
+                }
+                GroupRemoveUnit(shenGroup, unit);
+            }
+        }
+
+        /// <summary>
+        /// Changes the owner of all units inside the specified <paramref name="area"/> from <paramref name="owningPlayer"/> to <paramref name="newOwningPlayer"/> 
+        /// and renders all non-building units invisible.
+        /// </summary>
+        /// <param name="area"></param>
+        /// <param name="owningPlayer"></param>
+        /// <param name="newOwningPlayer"></param>
+        public static void ChangeUnitsOwningPlayer(this rect area, player owningPlayer, player newOwningPlayer)
+        {
+            foreach (var unit in new GroupWrapper().EnumUnitsInRect(area).EmptyToList())
+            {
+                if (unit.OwningPlayer() == owningPlayer)
+                {
+                    unit.SetOwner(newOwningPlayer);
+                    if (!IsUnitType(unit, UNIT_TYPE_STRUCTURE))
+                    {
+                        unit.Show(false);
+                    }
+                }
+            }
+        }
+    }
+}
--- a/src/MacroTools/Extensions/RectExtensions.cs
+++ b/src/MacroTools/Extensions/RectExtensions.cs
@@ -1,20 +1,25 @@
 ﻿using MacroTools.Wrappers;
+using System.Collections.Generic;
 using static War3Api.Common;
 
 namespace MacroTools.Extensions
 {
     /// <summary>
-    /// Provides a helpful set of extension methods for <see cref="rect"/>s.
+    /// Provides a helpful set of extension methods for native Warcraft 3 <see cref="rect"/>s.
     /// </summary>
     public static class RectExtensions
     {
         /// <summary>
-        /// Renders all units inside the specified area invulnerable/>
+        /// Renders all units inside the specified <paramref name="area"/> that belong to <paramref name="owningUnitsPlayer"/> invulnerable/>
         /// </summary>
         /// <param name="area"></param>
-        public static void MakeUnitsInvulnerable(this rect area)
-        {
+        /// <param name="owningUnitsPlayer"></param>
+        /// <returns>A list of all units found in the specified area that belong to <paramref name="owningUnitsPlayer"/></returns>
+        // Todo: Rename to something more sensible
+        public static List<unit> MakeUnitsInvulnerable(this rect area, player owningUnitsPlayer)
+        {        
             var shenGroup = CreateGroup();
+            List<unit> groupUnits = new();
             GroupEnumUnitsInRect(shenGroup, area, null);
             bool execute = true;
             while (execute)
@@ -25,12 +30,14 @@ namespace MacroTools.Extensions
                     execute = false;
                     continue;
                 }
-                if (GetOwningPlayer(unit) == Player(GetPlayerNeutralPassive()))
+                if (GetOwningPlayer(unit) == owningUnitsPlayer)
                 {
                     SetUnitInvulnerable(unit, true);
+                    groupUnits.Add(unit);
                 }
-                GroupRemoveUnit(shenGroup, unit);
+                GroupRemoveUnit(shenGroup, unit);               
             }
+            return groupUnits;
         }
 
         /// <summary>
--- a/src/WarcraftLegacies.Source/Quests/Sentinels/QuestScepterOfTheQueenSentinels.cs
+++ b/src/WarcraftLegacies.Source/Quests/Sentinels/QuestScepterOfTheQueenSentinels.cs
@@ -2,6 +2,7 @@
 using MacroTools.FactionSystem;
 using MacroTools.QuestSystem;
 using MacroTools.QuestSystem.UtilityStructs;
+using System.Collections.Generic;
 using WarcraftLegacies.Source.Setup;
 using WarcraftLegacies.Source.Setup.Legends;
 using WCSharp.Shared.Data;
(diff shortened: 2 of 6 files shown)


src/MacroTools/Extensions/RectExtensions.cs
```diff
@@ -0,0 +1,58 @@
...
+    /// <summary>
+    /// Provides a helpful set of extension methods for <see cref="rect"/>s.
+    /// </summary>
+    public static class RectExtensions
+    {
+        /// <summary>
+        /// Renders all units inside the specified area invulnerable/>
+        /// </summary>
+        /// <param name="area"></param>
+        public static void MakeUnitsInvulnerable(this rect area)
```
YakaryBovine, line 15: I think we need a different method name for this since it does more than just make units invulnerable. But it's a good method, and I'll backport it to some of the older Quests I did. Not sure on a name yet...

src/WarcraftLegacies.Source/Quests/Sentinels/QuestScepterOfTheQueenSentinels.cs
```diff
@@ -0,0 +1,50 @@
...
+using MacroTools.QuestSystem;
+using MacroTools.QuestSystem.UtilityStructs;
+using WarcraftLegacies.Source.Setup;
+using WarcraftLegacies.Source.Setup.Legends;
+using WCSharp.Shared.Data;
+using static War3Api.Common;
+
+namespace WarcraftLegacies.Source.Quests.Sentinels
+{
+    public class QuestScepterOfTheQueenSentinels : QuestData
```
YakaryBovine, line 12: Let's make this `sealed`. We never want to inherit from it.

src/WarcraftLegacies.Source/Quests/Warsong/QuestScepterOfTheQueenWarsong.cs
```diff
@@ -0,0 +1,35 @@
...
+using MacroTools.QuestSystem;
+using MacroTools.QuestSystem.UtilityStructs;
+using WarcraftLegacies.Source.Setup;
+using WarcraftLegacies.Source.Setup.Legends;
+using WCSharp.Shared.Data;
+using static War3Api.Common;
+
+namespace WarcraftLegacies.Source.Quests.Warsong
+{
+    public class QuestScepterOfTheQueenWarsong : QuestData
```
YakaryBovine, line 12: Make `sealed`.

src/WarcraftLegacies.Source/Quests/Warsong/QuestScepterOfTheQueenWarsong.cs
```diff
@@ -0,0 +1,35 @@
...
+        public QuestScepterOfTheQueenWarsong(Rectangle area) : base("Royal Plunder", "Remnants of the ancient Highborne survive within the ruins of Dire Maul. If Feathermoon Stronghold falls, it would become a simple matter to slaughter the Highborne and plunder their artifacts.", "ReplaceableTextures\\CommandButtons\\BTNNagaWeaponUp2.blp")
+        {
+            HighBourneArea = area;
+            AddObjective(new ObjectiveLegendNotPermanentlyDead(LegendWarsong.StonemaulKeep));
+            AddObjective(new ObjectiveLegendDead(LegendSentinels.legendFeathermoon));
+            AddObjective(new ObjectiveAnyUnitInRect(HighBourneArea, "Dire Maul", true));
+            HighBourneArea.Rect.MakeUnitsInvulnerable();
+        }
+
+        private Rectangle HighBourneArea { get; }
```
YakaryBovine, line 23: My convention for private fields is `private Rectangle _highbourneArea`.

src/WarcraftLegacies.Source/Quests/Warsong/QuestScepterOfTheQueenWarsong.cs
```diff
@@ -0,0 +1,35 @@
...
+        protected override string CompletionPopup => "The Highborne are no longer implicitly defended by the Night Elven presence at Feathermoon Stronghold. The Horde unleashes their full might against these Night Elven arcanists.";
```
YakaryBovine, line 24: Add `/// <inheritdoc />` above this or the compiler will yell at me.

src/WarcraftLegacies.Source/Quests/Warsong/QuestScepterOfTheQueenWarsong.cs
```diff
@@ -0,0 +1,35 @@
...
+        protected override string RewardDescription => "Gain the Scepter of the Queen and turn all units in Dire Maul hostile";
```
YakaryBovine, line 25: `/// <inheritdoc />`

src/WarcraftLegacies.Source/Quests/Warsong/QuestScepterOfTheQueenWarsong.cs
```diff
@@ -0,0 +1,35 @@
...
+
+        protected override void OnComplete(Faction whichFaction)
```
YakaryBovine, line 27: `/// <inheritdoc />`

src/WarcraftLegacies.Source/Quests/Warsong/QuestScepterOfTheQueenWarsong.cs
```diff
@@ -0,0 +1,35 @@
...
+        {
+            base.OnComplete(whichFaction);
```
YakaryBovine, line 29: Do you think we need to call the base? If we do I might need to add this to every other quest.

src/WarcraftLegacies.Source/Quests/Warsong/QuestScepterOfTheQueenWarsong.cs
```diff
@@ -0,0 +1,35 @@
...
+{
+    public class QuestScepterOfTheQueenWarsong : QuestData
+    {
+        public QuestScepterOfTheQueenWarsong(Rectangle area) : base("Royal Plunder", "Remnants of the ancient Highborne survive within the ruins of Dire Maul. If Feathermoon Stronghold falls, it would become a simple matter to slaughter the Highborne and plunder their artifacts.", "ReplaceableTextures\\CommandButtons\\BTNNagaWeaponUp2.blp")
+        {
+            HighBourneArea = area;
+            AddObjective(new ObjectiveLegendNotPermanentlyDead(LegendWarsong.StonemaulKeep));
+            AddObjective(new ObjectiveLegendDead(LegendSentinels.legendFeathermoon));
+            AddObjective(new ObjectiveAnyUnitInRect(HighBourneArea, "Dire Maul", true));
+            HighBourneArea.Rect.MakeUnitsInvulnerable();
```
YakaryBovine, line 20: So I think this isn't quite the pattern we want. We should:
1. When the Quest is created, make all the units invulnerable and put them into a List<unit> field.
2. When the Quest is completed, iterate across that list and rescue the units inside.
3. When the Quest is failed, iterate across that list and rescue the units for neutral hostile.

The reason for this is so that we can avoid accidentally rescuing neutral passive units that may have been moved to the region after the game has already started. We don't want to rescue those.

src/WarcraftLegacies.Source/Quests/Sentinels/QuestScepterOfTheQueenSentinels.cs
```diff
@@ -0,0 +1,50 @@
...
+            HighBourneArea.Rect.MakeUnitsInvulnerable();
+        }
+
+        private Rectangle HighBourneArea { get; }
+        protected override string CompletionPopup => "The Shen'dralar, the Highborne survivors of the Sundering, swear allegiance to their fellow Night Elves. As a sign of their loyalty, they offer up an artifact they have guarded for thousands of years: the Scepter of the Queen.";
+        protected override string RewardDescription => "Gain the Scepter of the Queen and control of all units in Dire Maul";
+
+        protected override void OnComplete(Faction whichFaction)
+        {
+            ArtifactSetup.ArtifactScepterofthequeen?.Item.SetPosition(new Point(GetRectCenterX(HighBourneArea.Rect), GetRectCenterY(HighBourneArea.Rect)));
```
YakaryBovine, line 30: Nice simplification you can do here:
```suggestion
            ArtifactSetup.ArtifactScepterofthequeen?.Item.SetPosition(HighBourneArea.Center);
```

src/WarcraftLegacies.Source/Quests/Sentinels/QuestScepterOfTheQueenSentinels.cs
```diff
@@ -0,0 +1,50 @@
...
+{
+    public class QuestScepterOfTheQueenSentinels : QuestData
+    {
+        public QuestScepterOfTheQueenSentinels(Rectangle area) : base("Return to the Fold", "Remnants of the ancient Highborne survive within the ruins of Dire Maul. If Stonemaul falls, it would be safe for them to come out.", "ReplaceableTextures\\CommandButtons\\BTNNagaWeaponUp2.blp")
+        {
+            HighBourneArea = area;
+            AddObjective(new ObjectiveLegendNotPermanentlyDead(LegendSentinels.legendFeathermoon));
+            AddObjective(new ObjectiveLegendDead(LegendWarsong.StonemaulKeep));
+            AddObjective(new ObjectiveAnyUnitInRect(Regions.HighBourne, "Dire Maul", true));
+            ResearchId = FourCC("R020");
```
YakaryBovine, line 20:
```suggestion
            ResearchId = Constants.UPGRADE_R02O_QUEST_COMPLETED_RETURN_TO_THE_FOLD_SENTINELS;
```
We use constants extracted from the map now. You can find these by typing in Constants.R02O into your IDE and letting it autocomplete for you.

src/WarcraftLegacies.Source/Quests/Sentinels/QuestScepterOfTheQueenSentinels.cs
```diff
@@ -0,0 +1,50 @@
...
+            ArtifactSetup.ArtifactScepterofthequeen?.Item.SetPosition(new Point(GetRectCenterX(HighBourneArea.Rect), GetRectCenterY(HighBourneArea.Rect)));
+            SetPlayerTechResearched(whichFaction.Player, ResearchId, 1);
+            if (whichFaction.Player != null)
+                HighBourneArea.Rect.ChangeUnitsOwningPlayer(Player(GetPlayerNeutralPassive()), whichFaction.Player);
+        }
+
+        protected override void OnAdd(Faction whichFaction)
+        {
+            base.OnAdd(whichFaction);
+            whichFaction.ModObjectLimit(ResearchId, Faction.UNLIMITED);
```
YakaryBovine, line 39: You can delete this whole method; you've already registered the ResearchId in the constructor so this will get handled for you.

src/WarcraftLegacies.Source/Quests/Sentinels/QuestScepterOfTheQueenSentinels.cs
```diff
@@ -0,0 +1,50 @@
...
+        }
+
+        protected override void OnFail(Faction whichFaction)
+        {
+            base.OnFail(whichFaction);
+            whichFaction.ModObjectLimit(ResearchId, -Faction.UNLIMITED);
```
YakaryBovine, line 45: You can also get rid of this!

YakaryBovine (changes requested):
Amazing work, you've mostly nailed it as far as I can tell. More style comments incoming, plus a few functional ones this time!

src/WarcraftLegacies.Source/Quests/Warsong/QuestScepterOfTheQueenWarsong.cs
YakaryBovine: Sorry, scratch point 3. For this quest we can't give it to hostile on fail, since the rival player might still complete their version of the quest.

YakaryBovine:
By the way, I agree with you that extension methods don't necessarily feel right here. I'm OK with it though since I can't think of anything much better.

src/WarcraftLegacies.Source/Quests/Warsong/QuestScepterOfTheQueenWarsong.cs
TracyTheBoy: No the call is not neccessary since the base method is not doing anything. I removed it.
If there are no cases where the virtual methods in QuestData have default behaviour they could all be made abstract to enforce overrides in child classes

YakaryBovine: Cool, makes sense. I can guarantee the base methods will never do anything.
At the same time I don't think we want to make them `abstract` since we don't really want to force behaviour override. E.g. there are many quests which don't do anything on failure, so why enforce writing something for that method? Up for debate.

src/WarcraftLegacies.Source/Quests/Warsong/QuestScepterOfTheQueenWarsong.cs
TracyTheBoy: To Addresss this, `MakeUnitsInvulnerable` now returns the units found.
And then I added another Method to `PlayerExtensions` that takes a List of units but does basically the same as 
`ChangeUnitsOwngingPlayer` in rect extensions.
I did this because the rect is no longer needed there

YakaryBovine (approved):
Well done!

Issue #87, opened 2022-09-30
Kul'tiras quests


YakaryBovine:
Implement all Kul'tiras quests from the live vJASS version.
Many of these have been started, but none of them have been tested. They will likely need changes.

TracyTheBoy:
Hi, I am working on these quests right now and would like to test my changes but I am having issues starting the map as Kul'Tiras.
The game crashes during the loading screen when I set my TestingPlayerSlot in the appsettings to 21 which is Kul'Tiras as far as I can tell.
I tested it with other player numbers I used previously and those are still working so it must be an issue specific to that faction

YakaryBovine:
Sweet, I'll check in a bit when I can get a break from work. My guess would be that Kul'tiras is currently set to the wrong player slot entirely in-game and the one I'm using is actually unused. But we'll see!

TracyTheBoy:
Hi, that was indeed it. I compared the player slots in the game with the configurations in the code and they indeed didn't fit.
I matched them and now it's working.
Thanks

YakaryBovine:
Ah, brilliant!

TracyTheBoy:
So I got this mostly working but I am facing some issues I don't know how to solve on that last quest `QuestUnlockShip`.
When the quest is failed your heroes are supposed to spawn inside the ship but I can't get them to teleport to the correct location.
When i use the same location `ShipAmbient` as in the Jass version, the heroes spawn in a completely different location.
Same thing with the ship. It's supposed to spawn at the coast of Southshore but it spawns too far to the right even though I am using the same location as in the Jass Verison.
The coordinates in the `Regions` class probably don't match the actual locations in the map but not sure.
Anyways I'm kinda lost on how to deal with this properly

YakaryBovine:
Sweet, I'll look into it tomorrow my time. Sounds like some region stupidity I'll have to solve in the map file; your code is probably fine.

YakaryBovine:
Ok - the ShipAmbient region was in a really odd place so I moved it. Not sure if it was the correct move but we'll find out. I didn't move SouthshoreUnlock; it's needed for another quest, and I think yours should work anyway since you're not spawning the ship, only issuing an order to it.

Unfortunately the Regions.cs file is generated from the actual in-map Regions list: 
![image](https://user-images.githubusercontent.com/8995690/200701895-bcb7e638-460c-4286-8fe5-739f874cfea6.png)
So you would need to edit the actual .w3x file to change them.

PR #141, opened 2022-11-09
#87 Kul'Tiras quests

TracyTheBoy:
This bit in `QuestTheramore` could be simplified if all theramore units would start as neutral passive by default. 
Definetly should be kept in mind because this will stop working if someone decides to 'fix' that in the map itself.
`   if (KultirasSetup.Kultiras?.Player != null)
      _rescueUnits = theramoreRect.PrepareUnitsForRescue(KultirasSetup.Kultiras.Player); 
      Player(PLAYER_NEUTRAL_PASSIVE).RescueGroup(_rescueUnits); // Have to make units neutral because they are owned by Kul'Tiras by default`

Also english is not my first language so I apologize for any spelling and grammar mistakes I made in the flavour texts and comments

### DIFF
--- a/src/WarcraftLegacies.Source/Quests/KulTiras/QuestUnlockShip.cs
+++ b/src/WarcraftLegacies.Source/Quests/KulTiras/QuestUnlockShip.cs
@@ -1,20 +1,30 @@
-using System.Collections.Generic;
+﻿using System.Collections.Generic;
 using MacroTools.Extensions;
 using MacroTools.FactionSystem;
 using MacroTools.QuestSystem;
 using MacroTools.QuestSystem.UtilityStructs;
-using MacroTools.Wrappers;
 using WarcraftLegacies.Source.Setup.Legends;
 using WCSharp.Shared.Data;
 using static War3Api.Common;
+using static War3Api.Blizzard;
+using WarcraftLegacies.Source.Setup.FactionSetup;
+using WarcraftLegacies.Source.Setup;
 
 namespace WarcraftLegacies.Source.Quests.KulTiras
 {
+  /// <summary>
+  /// Proudmoore captial ship starts locked. Boralus and Dazar'Alor must be captured to unlock it.
+  /// </summary>
   public sealed class QuestUnlockShip : QuestData
   {
     private readonly unit _proudmooreCapitalShip;
     private readonly List<unit> _rescueUnits = new();
-
+    private readonly Rectangle _rescueRect;
+    /// <summary>
+    /// Initializes a new instance of the <see cref="QuestUnlockShip"/> class.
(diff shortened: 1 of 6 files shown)


YakaryBovine:
Good call, I'll do that. Made a ticket here: https://github.com/AzerothWarsLR/WarcraftLegacies/issues/143

src/WarcraftLegacies.Source/Quests/KulTiras/QuestUnlockShip.cs
```diff
@@ -1,66 +1,98 @@
...
+        completingFaction.Player.RescueGroup(_rescueUnits);
+        completingFaction.Player.AddGold(500);
+        completingFaction.Player.AddLumber(2000);
+        _proudmooreCapitalShip.Rescue(completingFaction.Player);
+      }
+      else
+      {
+        _proudmooreCapitalShip.Rescue(Player(PLAYER_NEUTRAL_AGGRESSIVE));
+      }
+      PauseUnitBJ(false, _proudmooreCapitalShip);
```
YakaryBovine, line 92: We should avoid using BJ functions where possible since they're sometimes very inefficient. Try:
```suggestion
      _proudmooreCapitalShip.Pause(false);
```
If that extension method doesn't exist, try:
```suggestion
      PauseUnit(_proudmooreCapitalShip, false);
```

src/WarcraftLegacies.Source/Quests/KulTiras/QuestUnlockShip.cs
```diff
@@ -1,66 +1,98 @@
...
+      IssuePointOrderLocBJ(_proudmooreCapitalShip, "move", GetRectCenter(Regions.SouthshoreUnlock.Rect));
```
YakaryBovine, line 93:
```suggestion
      IssuePointOrder(_proudmooreCapitalShip, "move", Regions.SouthshoreUnlock.Center.X, Regions.SouthshoreUnlock.Center.Y);
```

YakaryBovine (changes requested):
Great stuff. Your new extension methods have really made quest code a lot simpler.

src/WarcraftLegacies.Source/Quests/KulTiras/QuestUnlockShip.cs
TracyTheBoy: I see.
It's hard to decide which of all the different methods to use when they all sound like they're doing bascially the same thing. 
But I see that you're opting for the exentsion methods wherever possible so I'll try to keep that in mind going forward

YakaryBovine (approved):
Perfect

PR #155, opened 2022-11-10
Victory Imminent Message Spam Fix

SavanticIO:
Get the former owner team of the changing control point and check if the new team is different from the former before doing victory checking.

This should sort the issue where messages get spammed on the screen when a player leaves the game whose team have over the CPS_WARNING number of control points.

Fixed an index out of bounds exception when trying to distribute units to other players in the faction


YakaryBovine (approved):
Looks perfect to me.

Issue #183, opened 2022-11-11
No Timer and turn counter display


SebastianTanoni:

SebastianTanoni:
![image](https://user-images.githubusercontent.com/47048283/201360173-81227513-ab18-43be-a90d-cbbe4d6b826c.png)

SavanticIO:
This is a super easy one unless the plan is to redo the timer system. @YakaryBovine If the plan is just to get this one working again then I'll grab this one.

YakaryBovine:
@SavanticIO Yep, sounds great.

PR #273, opened 2022-11-11
no timer and turn counter display. Issue #183

SavanticIO:
Fix for issue #183

Just added calling GameTime.Setup() during the setup process


YakaryBovine (approved):
That's hilarious. Good find. That's what I get for starting this whole thing off with a hacked together transpiler!

Issue #244, opened 2022-11-11
When a player leaves, all his units get refunded and deleted


SebastianTanoni:

SebastianTanoni:
as opposed to only certain units, like elites and heroes, get removed and refunded

SavanticIO:
@YakaryBovine I'll take this one

SavanticIO:
I'm not sure this is an issue. As far as I can tell when a player leaves, all their units are refunded and their buildings are given to the remaining team members

YakaryBovine:
@SavanticIO Really? I don't remember doing this... I'll check it out now.

YakaryBovine:
@SavanticIO You are completely correct. Closing this is as it's already been done.

PR #274, opened 2022-11-12
when a player leaves all his units get refunded and deleted

SavanticIO:
Fix for issue #244 

This might not be a complete list of 'special' units that require a refund.

Basically I got a list of all the refund units from the vJASS version and used regex to grab all the constants values.

If any more units are added that require a refund then a new config entry will need to be added. Would be nicer if there was a more programatic way to do this.


YakaryBovine:
Assuming I'm interpreting the requirement from @SebastianTanoni correctly...
I think you can just remove the Refund property entirely and refund _all_ non-structure units, with no other condition.

SavanticIO:
Might need clarification on the issue. Currently it does refund all non-structure units. My interpretation was that the ordinary units being removed was the issue.

Issue #144, opened 2022-11-09
Use new rescue-related extension methods across all relevant quests


YakaryBovine:
https://github.com/AzerothWarsLR/WarcraftLegacies/pull/139 introduced a useful set of extension methods for dealing with preparing bases for rescue in quests, and for rescuing those bases later. We have:
`RectangleExtensions.PrepareForRescue` to make a group of units invulnerable and put them into a collection.
and
`PlayerExtensions.RescueGroup` to rescue those units later when the quest is over.

This ticket is to ensure that these methods are used throughout Quest code where appropriate. E.g. this highlighted block of code in QuestKilsorrow.cs could be replaced with `PrepareForRescue`:
![image](https://user-images.githubusercontent.com/8995690/200737810-2c979d43-dad1-43fc-95e0-7254fbb27a78.png)

YakaryBovine:
Closing as stale.

PR #287, opened 2022-11-13
Frostwolf Quests and Mechanics

### LINKED ISSUE #184: Cannot rescue the trolls and orcs on Frostwolf Starting island
Completing Riders of the Storm is meant to give you control of every unit on Frostwolf's starting island, _but_ it should also be possible to rescue the groups of Trolls and Orcs just by approaching them, like in the campaign.

As far as I recall this is implemented in GUI in the vJASS version.

Frostwolf 1

### LINKED ISSUE #186: Completing Riders of the storm dropped me in a random spot, instead of near tanaris
image

### LINKED ISSUE #188: After riders on the storm, did not gain control of the taurens and pack kodo here
image

### LINKED ISSUE #189: Long march completion text is wrong
image

### LINKED ISSUE #192: Implement the Frostwolf Clan's Pack Kodo mechanic from the live vJASS version of the game

SavanticIO:
Fixed some issues around some of the starting Frostwolf quests.

Added the pack structures mechanic

I tried to make the structure packing mechanic generic enough that in theory it could be used for other buildings and units.

Closes #184 
Closes #186 
Closes #188 
Closes #189 
Closes #192


YakaryBovine:
Awesome work. I'll give this a proper review tomorrow but it looks great. I like the way you grouped the related tickets together like this as well.

FYI this doesn't actually cover #184. I'll update the description on that ticket so it's clearer what the issue is.

SavanticIO:
Unless I am completely misreading the requirements I think it should cover #184. 
I added a trigger for a unit belonging to the frostwolf faction entering that rect and then rescuing the passive units in the rect to the Sea Witch quest. While not technicly part of the quest, I wasn't sure where to put loose triggers and it also means that the trigger can be destroyed on that quest fail/complete as there is no need to be firing that event after that.

YakaryBovine:
Ah yes you're correct, my mistake.

src/MacroTools/Extensions/RectangleExtensions.cs
```diff
@@ -43,5 +43,22 @@ public static List<unit> PrepareUnitsForRescue(this Rectangle area, player ownin
       }
       return group;
     }
+    
+    /// <summary>
+    /// Sets the owner of all units inside the specified <paramref name="area"/> to the specifed <see cref="player" />
+    /// </summary>
+    /// <param name="area"></param> The area the units are in
+    /// <param name="whichPlayer"></param> The player to gain control of the units
+    public static void RescueNeutralUnitsInRect(this Rectangle area, player whichPlayer)
```
YakaryBovine, line 52: I've been thinking about this, and I think we should remove it and adopt the pattern @TracyTheBoy introduced for every Quest of this nature. Check out this ticket: https://github.com/AzerothWarsLR/WarcraftLegacies/issues/144

Basically, whenever a Quest with a rescue area is constructed, it should grab all of the units in that area and put them into a group. Then when the Quest is completed, it should rescue that group instead of rescuing units in the area. This avoids the possibility of us accidentally rescuing units that wander into the area later on.

src/MacroTools/Mechanics/StructurePacking.cs
```diff
@@ -0,0 +1,91 @@
...
+{
+  /// <summary>
+  /// System to allow structures to convert themselves into units. Those units can then use an ability to rebuild the building.
+  /// </summary>
+  public class StructurePacking
+  {
+    /// <summary>
+    /// Object that contains information and methods for packable structures
+    /// </summary>
+    public class PackableStructure
```
YakaryBovine, line 16:
```suggestion
    public sealed class PackableStructure
```
I'm personally a fan of making classes sealed when they're not explicitly designed to be inheritable.

src/MacroTools/Mechanics/StructurePacking.cs
```diff
@@ -0,0 +1,91 @@
...
+      private string? _structureModel;
+
+      /// <summary>
+      /// Method for creating new packable structure entries
+      /// </summary>
+      /// <param name="structureId"></param>
+      /// <param name="packedUnitId"></param>
+      /// <param name="buildAbility"></param>
+      /// <param name="structureModel"></param>
+      public static void Create(int structureId, int packedUnitId, int buildAbility, string? structureModel)
```
YakaryBovine, line 30: Could we do this for consistency with some of the other systems?
```suggestion
      public static void Register(int structureId, int packedUnitId, int buildAbility, string? structureModel)
```

src/MacroTools/Mechanics/StructurePacking.cs
```diff
@@ -0,0 +1,91 @@
...
+  {
+    /// <summary>
+    /// Object that contains information and methods for packable structures
+    /// </summary>
+    public class PackableStructure
+    {
+      private static readonly Dictionary<int, PackableStructure> PackableStructureById = new();
+      private int _packedUnitId;
+      private int _buildAbility;
+      private string? _structureModel;
```
YakaryBovine, line 21: Rather than making this nullable let's just initialize it to something valid.
```suggestion
      private string _structureModel = string.Empty;
```

src/WarcraftLegacies.Source/Mechanics/Frostwolf/FrostwolfStructurePackingConfig.cs
```diff
@@ -0,0 +1,29 @@
+﻿using System;
+using MacroTools.Mechanics;
+
+namespace WarcraftLegacies.Source.Mechanics.Frostwolf
+{
+  /// <summary>
+  /// Config file for setting up the packable structures for the Frostwolf faction
+  /// </summary>
+  public class FrostwolftructurePackingConfig : StructurePacking
```
YakaryBovine, line 9:
```suggestion
  public static class FrostwolfStrcturePackingConfig
```

src/WarcraftLegacies.Source/Quests/Frostwolf/QuestThunderBluff.cs
```diff
@@ -35,7 +35,7 @@ public sealed class QuestThunderBluff : QuestData
     }
 
     //todo: bad flavour
-    protected override string CompletionPopup => "The Crossroads have been constructed.";
+    protected override string CompletionPopup => "The Long March of the Tauren clans has ended and they have joined forces with the Horde.";
```
YakaryBovine, line 38: Nice, I like the flavour a lot.
```suggestion
    protected override string CompletionPopup => "The long march of the Tauren clans has ended, and they have joined forces with the Horde.";
```

src/WarcraftLegacies.Source/Quests/Frostwolf/QuestSeaWitch.cs
```diff
@@ -29,13 +30,15 @@ public sealed class QuestSeaWitch : QuestData
       AddObjective(new ObjectiveKillUnit(LegendNeutral.LegendSeawitch.Unit));
       AddObjective(new ObjectiveExpire(600));
       ResearchId = FourCC("R05H");
-
-      foreach (var unit in new GroupWrapper().EnumUnitsInRect(rescueRect).EmptyToList())
-        if (GetOwningPlayer(unit) == Player(PLAYER_NEUTRAL_PASSIVE))
+      _rescueUnits = rescueRect.PrepareUnitsForRescue(Player(PLAYER_NEUTRAL_PASSIVE));
+      CreateTrigger()
```
YakaryBovine, line 34: I think you meant to assign this to `_rescueTrigger`? Brilliant way to do it, anyway.

src/MacroTools/Mechanics/StructurePacking.cs
```diff
@@ -0,0 +1,91 @@
...
+      {
+        var packable = new PackableStructure
+        {
+          _packedUnitId = packedUnitId,
+          _buildAbility = buildAbility,
+          _structureModel = structureModel
+        };
+        PackableStructureById[structureId] = packable;
+        
+        PlayerUnitEvents.Register(PlayerUnitEvent.UnitTypeFinishesTraining, GetPackableStructureById(structureId).OnTrainUnitType,structureId);
```
YakaryBovine, line 40: So I suspect we can simplify this whole trigger massively through the use of anonymous functions. We didn't have those in the vJASS version you looked at so we had to use the Dictionary technique. Let's chat about it on Discord some time.

YakaryBovine (changes requested):
Awesome stuff, I like the way you think. You should add yourself to the Credits some time!

SavanticIO:
@YakaryBovine Using a lambda expression instead of maintaing a dictionary worked like a charm

YakaryBovine (approved):
Absolutely fantastic. Yeah, it's way cleaner without the dictionary!

PR #290, opened 2022-11-13
issues #190 #208 #241

### LINKED ISSUE #190: Wierd text for the tomb of sargeras quest
image

### LINKED ISSUE #208: Draktharon keep completion text wrong
image

### LINKED ISSUE #241: Stormwind quest requirement says town hall to townhal
image

Should be a castle

TracyTheBoy:
Wierd text for the tomb of sargeras quest #190
Draktharon keep completion text wrong #208
Stormwind quest requirement says town hall to townhal #241

Closes #190 
Closes #208 
Closes #241

### DIFF
--- a/src/MacroTools/QuestSystem/UtilityStructs/ObjectiveHeroWithLevelReachRect.cs
+++ b/src/MacroTools/QuestSystem/UtilityStructs/ObjectiveHeroWithLevelReachRect.cs
@@ -16,7 +16,7 @@ namespace MacroTools.QuestSystem.UtilityStructs
       _targetRect = targetRect.Rect;
       _targetLevel = targetLevel;
       region target = RectToRegion(_targetRect);
-      Description = $"Any hero reaches {rectName}";
+      Description = $"Any hero that is atleast level {targetLevel} reaches {rectName}";
       TriggerRegisterEnterRegion(_entersRect.Trigger, target, null);
       TriggerAddAction(_entersRect.Trigger, OnRegionEnter);
       PingPath = "MinimapQuestTurnIn";
--- a/src/MacroTools/QuestSystem/UtilityStructs/ObjectiveHeroWithLevelReachRect.cs
+++ b/src/MacroTools/QuestSystem/UtilityStructs/ObjectiveHeroWithLevelReachRect.cs
@@ -1,9 +1,13 @@
-﻿using MacroTools.Wrappers;
+﻿using MacroTools.Extensions;
+using MacroTools.Wrappers;
 using WCSharp.Shared.Data;
 using static War3Api.Common;
 
 namespace MacroTools.QuestSystem.UtilityStructs
 {
+  /// <summary>
+  /// Defines an <see cref="Objective"/> that completes when any hero of a certain level or higher enters a specific region.
+  /// </summary>
   public sealed class ObjectiveHeroWithLevelReachRect : Objective
   {
     private readonly TriggerWrapper _entersRect = new();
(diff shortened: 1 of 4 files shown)


src/MacroTools/QuestSystem/UtilityStructs/ObjectiveHeroWithLevelReachRect.cs
```diff
@@ -16,7 +16,7 @@ public ObjectiveHeroWithLevelReachRect(int targetLevel, Rectangle targetRect, st
       _targetRect = targetRect.Rect;
       _targetLevel = targetLevel;
       region target = RectToRegion(_targetRect);
-      Description = $"Any hero reaches {rectName}";
+      Description = $"Any hero that is atleast level {targetLevel} reaches {rectName}";
```
YakaryBovine, line 19:
```suggestion
      Description = $"Any level {targetLevel}+ hero reaches {rectName}";
```
This is a bit shorter and we have very limited space at times.

src/WarcraftLegacies.Source/Quests/Stormwind/QuestStormwindCity.cs
```diff
@@ -1,62 +1,71 @@
...
+
+      if (GetLocalPlayer() == completingFaction.Player) 
+        PlayThematicMusic("war3mapImported\\StormwindTheme.mp3");
     }
 
+    /// <inheritdoc />
     protected override void OnAdd(Faction whichFaction)
     {
-      whichFaction.ModObjectLimit(QuestResearchId, 1);
+      whichFaction.ModObjectLimit(ResearchId, 1);
```
YakaryBovine, line 68: You can actually delete this entire method, as the function is already called by virtue of it being set as ResearchId.

src/WarcraftLegacies.Source/Quests/QuestTombOfSargeras.cs
```diff
@@ -26,18 +30,15 @@ public sealed class QuestTombOfSargeras : QuestData
...
+      Global = true;
     }
     
     /// <inheritdoc />
     protected override string RewardDescription => "The Tomb of Sargeras opens";
 
-    //Todo; it would be really cool if this depended on the hero that opens it
     /// <inheritdoc />
-    protected override string CompletionPopup => "The Tomb of Sargeras has been opened.";
+    protected override string CompletionPopup => $"The Tomb of Sargeras has been opened by {GetTriggerUnit().GetProperName()}";
```
YakaryBovine, line 41: This is brilliant, but is also a slightly risky way to do it. We cannot guarantee that the last Objective completed is ObjectiveHeroWithLevelReachRect, so we might not have a valid TriggerUnit(). 
Instead I would suggest modifying ObjectiveHeroWithLevelReachRect to set a public property on itself whenever it gets completed, then use that public property here.

YakaryBovine (changes requested):
Brilliant stuff as always.

src/WarcraftLegacies.Source/Quests/QuestTombOfSargeras.cs
TracyTheBoy: I hope I solved that the way you wanted to.
I added `public string EnteredRegionUnitName { get; private set; } = "an unknown hero.";` inside the `ObjectiveHeroWithLevelReachRect` and then assigned that in the `private void OnRegionEnter()` 
I then added a private field to `QuestTombOfSargeras` which stores the objective so I can use the name in the `CompletionPopup`

YakaryBovine (approved):
Excellent!

Issue #279, opened 2022-11-12
Rescue Unit function Add a boolean to hide buildings too 


SebastianTanoni:
make it so i can decide wether i want buildings to be hiden or not in the preparerescue function

SavanticIO:
@YakaryBovine I'll be implimenting this and a few other overload methods to PrepareUnitsForRescue as a consequence of fixing #249

YakaryBovine:
Awesome, that makes sense to me.

Issue #251, opened 2022-11-11
Control points in outland should turn hostile when i leave (Draenei)


SebastianTanoni:
![image](https://user-images.githubusercontent.com/47048283/201410705-85a209f0-ac5e-4ed7-a848-9fb57723aa27.png)

SavanticIO:
@YakaryBovine

Issue #250, opened 2022-11-11
Broken civilisation didnt fail when i left early


SebastianTanoni:
![image](https://user-images.githubusercontent.com/47048283/201410624-85dabbc3-f40c-427c-ba13-9a8e898fe2a7.png)
Exodar is destroyed right now

SavanticIO:
@YakaryBovine

Issue #249, opened 2022-11-11
QuestWarnBase should make the Draenei bases neutral passive and vulnerable, not neutral passive and invulnerable


SebastianTanoni:

YakaryBovine:
@SebastianTanoni Could you clarify how this is meant to work? Are the bases just meant to start Neutral Passive and _vulnerable_?

SebastianTanoni:
> Correct, since if draenei is too slow, Fel Horde could destroy them still. If not, draenei could wait forever without unlocking them to lock fel horde

SavanticIO:
@YakaryBovine looks like there are a few Draenei related issues. I'll take a look at those.

PR #312, opened 2022-11-15
5 draenei fixes

SavanticIO:
Changes for issues;
Closes https://github.com/AzerothWarsLR/WarcraftLegacies/issues/249
Closes https://github.com/AzerothWarsLR/WarcraftLegacies/issues/250
Closes https://github.com/AzerothWarsLR/WarcraftLegacies/issues/251
Closes https://github.com/AzerothWarsLR/WarcraftLegacies/issues/222
Closes https://github.com/AzerothWarsLR/WarcraftLegacies/issues/279


src/MacroTools/Extensions/RectangleExtensions.cs
```diff
@@ -21,24 +21,92 @@ public static void AddSound(this Rectangle region, sound soundHandle)
...
     }
-
+    
     /// <summary>
-    /// Renders all units inside the specified <paramref name="area"/> that belong to <paramref name="owningUnitsPlayer"/> invulnerable.
+    /// Creates a group of units inside the specified <paramref name="area"/> that belong to <paramref name="owningUnitsPlayer"/>.
+    /// </summary>
+    /// <param name="area">The area in which to get the group of units.</param>
+    /// <returns>A list of all units found in the specified area</returns>
+    public static List<unit> PrepareUnitsForRescue(this Rectangle area)
```
YakaryBovine, line 30: I think what you've with these overloads is great, but I think the API could be cleaner.
Ultimately we only have a few combinations of valid inputs to use. I think instead of having a bunch of parameters, we should just have two: the area, and an Enum that indicates the rescue "mode". The mode would dictate the behaviour. What do you think?

YakaryBovine (changes requested):
Fantastic stuff. I made one investigatory comment, and everything else looks perfect.

src/MacroTools/Extensions/RectangleExtensions.cs
SavanticIO: I think it is a fantastic way to clean the messiness up

YakaryBovine (approved):
Brilliant!

Issue #352, opened 2022-11-16
I killed Morghor but it did not register the quest objective


SebastianTanoni:
![image](https://user-images.githubusercontent.com/47048283/202268585-51e55cd1-064d-456d-892c-00a095815534.png)

I must have messed up somewhere but i cant find where

SavanticIO:
@YakaryBovine I'll take this

YakaryBovine:
Closed by https://github.com/AzerothWarsLR/WarcraftLegacies/pull/367.

Issue #255, opened 2022-11-11
Make Illidan starting base change to his control after the neutrals have turned hostile


SebastianTanoni:

YakaryBovine:
@SebastianTanoni Need a little more detail. When do the neutrals turn hostile? Is this part of a quest?

SebastianTanoni:
A bunch of neutrals around the map start as player 16. Because i overload the neutrals hostile and passive.

So when the game starts, it picked every unit of player 15? whatever player number illidan is, and turned them neutral hostile.

After that is done, his starting base (ilidan's) need to be turned back to his control so he can play the game

YakaryBovine:
Awesome, thankyou.

SavanticIO:
@YakaryBovine I'll take this

SavanticIO:
@SebastianTanoni What region is Illidan's starting base in? None of the names in the regions file make it super obvious

SebastianTanoni:
@SavanticIO there might be no region, so someone will have to place it :) il do it in about 2h15!

SebastianTanoni:
Name of the region: IllidanStartingPosition

YakaryBovine:
Closed by https://github.com/AzerothWarsLR/WarcraftLegacies/pull/367.

SebastianTanoni:
This was not fixed

YakaryBovine:
Closed by https://github.com/AzerothWarsLR/WarcraftLegacies/commit/81dee5de0fa04eaa0d78f5707430589567c70635.

Issue #226, opened 2022-11-11
Destroying Stormwind keep while it is neutral promps legendary creep death message


SebastianTanoni:
![image](https://user-images.githubusercontent.com/47048283/201395321-1c8ed3ad-4d9e-4da8-ba98-add49513a5f7.png)

SavanticIO:
@YakaryBovine I'll take this one.

@SebastianTanoni Should the 'LEGENDARY CREEP DEATH' message only appear if the unit is NEUTRAL_AGGRESSIVE?

SebastianTanoni:
@SavanticIO yes, but more precily, it will never trigger on a building. Legendary creeps are always units or heroes

YakaryBovine:
Closed by #367.

PR #367, opened 2022-11-17
8 various bugs

SavanticIO:
#226 - Changed so the 'LEGENDARY CREEP DEATH' message only displays for hostile non-structure units

#255 - Created a prepared rescue group for illidans starting region and then rescued the group to illidan faction

#352 - Mor'ghor is a hero so needs the PermaDies = true flag

The Legion quest 'Infernal Foothold' had the wrong upgrade requirments for one of its objectives


SavanticIO:
The 'IllidanStartingPosition' Isn't in the Regions.cs file yet so will need the constants generate to be run and those files updated. Didn't seem like my place to update those as some one who doesn't edit the map.

YakaryBovine:
Great finds and fixes, thankyou!

Issue #179, opened 2022-11-11
Custom frames: Books should reshuffle their contents whenever an item is removed from them


YakaryBovine:
In https://github.com/AzerothWarsLR/WarcraftLegacies/pull/135, we made it so that Artifact Books shuffle their contents to fit the available space when an Artifact is removed from the game. We should move that logic to the generic Book/Page/Card classes so that it works for all other Books. PowerBook is of particular interest.

PR #384, opened 2022-11-19
Book related issues #179  #281

TracyTheBoy:
Implemented the same logic to re-render the contents of the `PowerBook` as for the `ArtiFactBook`.
Also added `OrderBy` extension methos to the Get methods of powers and artifacts.

I looked at your commit for #372 and a bug we've already fixed in #135 came back with that commit. 
Basically when an item gets removed so that the page is now empty, even after reshuffling the book it shows the wrong page because `_activePageIndex` does not get reset.
So I fixed that.

I realise that with the way I fixed this I kinda reverted what you did in #372 and I am not too happy with the way this turned out but I couldn't think of a better solution so yea.

### DIFF
--- a/src/MacroTools/ArtifactSystem/ArtifactManager.cs
+++ b/src/MacroTools/ArtifactSystem/ArtifactManager.cs
@@ -1,5 +1,6 @@
 ﻿using System;
 using System.Collections.Generic;
+using System.Linq;
 using MacroTools.Extensions;
 using MacroTools.ShoreSystem;
 using WCSharp.Events;
@@ -62,7 +63,7 @@ namespace MacroTools.ArtifactSystem
     /// Returns all <see cref="Artifact"/>s registered to the system.
     /// </summary>
     public static IEnumerable<Artifact> GetAllArtifacts() =>
-      AllArtifacts.AsReadOnly();
+      AllArtifacts.AsReadOnly().OrderBy(a => GetItemName(a.Item));
 
     /// <summary>
     /// Completely removes the given Artifact from the game.
--- a/src/MacroTools/BookSystem/BookManager.cs
+++ b/src/MacroTools/BookSystem/BookManager.cs
@@ -1,7 +1,5 @@
 ﻿using System;
 using System.Collections.Generic;
-using MacroTools.Extensions;
-using MacroTools.Frames;
 using static War3Api.Common;
 
 namespace MacroTools.BookSystem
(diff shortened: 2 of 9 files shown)


YakaryBovine:
Interesting! Looks pretty good to me.
The changes I made in #372 were intended to stop a desync issue with the way we were handling Books, so I'll just have to make sure that hasn't come back. But otherwise I think this might be an improvement. Will give it a proper review when I can.

src/MacroTools/ArtifactSystem/ArtifactManager.cs
```diff
@@ -62,7 +63,7 @@ public static void Register(Artifact artifact)
     /// Returns all <see cref="Artifact"/>s registered to the system.
     /// </summary>
     public static IEnumerable<Artifact> GetAllArtifacts() =>
-      AllArtifacts.AsReadOnly();
+      AllArtifacts.AsReadOnly().OrderBy(a => GetItemName(a.Item));
```
YakaryBovine, line 66: Great change.

src/MacroTools/BookSystem/Powers/PowerBook.cs
```diff
@@ -13,18 +14,17 @@ public sealed class PowerBook : Book<PowerPage>
...
-    
+
     /// <summary>
     /// Initializes a new instance of the <see cref="PowerBook"/> class.
     /// </summary>
     /// <param name="trackedPlayer">The player to show <see cref="Power"/>s for.</param>
-    public PowerBook(player trackedPlayer) : base(0.3f, 0.39f, 0.02f, 0.015f)
+    public PowerBook(player trackedPlayer) : base(0.3f, 0.39f, 0.02f, 0.015f, "Powers", new Point(0.4f, 0.36f), 
+      BlzGetFrameByName("UpperButtonBarMenuButton", 0), 
+      trackedPlayer == null || trackedPlayer == GetLocalPlayer())
```
YakaryBovine, line 24: This is the only bit I'm concerned about. It's _really_ unclear just from the constructor why this is necessary. I only understand it because I have so much context.

I'm not sure what the solution is but we need to up the readability here. The penalty for getting this wrong is a completely broken game and I don't trust myself to maintain this pattern for future Books.

src/MacroTools/FactionSystem/Faction.cs
```diff
@@ -518,7 +519,8 @@ public void Obliterate()
     /// </summary>
     public IEnumerable<Power> GetAllPowers()
     {
-      foreach (var power in _powers) yield return power;
+      foreach (var power in _powers.OrderBy(p => p.Name))
```
YakaryBovine, line 522: I actually don't think we want to sort this one by name. Sorting it by order of acquirement is better.

src/MacroTools/BookSystem/BookManager.cs
```diff
@@ -18,32 +16,9 @@ public static class BookManager
     /// Registers a <see cref="IBook"/> as being visible to all players.
     /// </summary>
     /// <param name="book">The book to register.</param>
-    /// <param name="whichPlayer">If specified, the Book can only be seen by this player.</param>
-    public static void Register(IBook book, player? whichPlayer = null)
+    public static void Register(IBook book)
     {
       Books.Add(book);
-      var launcherButton = new Button("ScriptDialogButton", BlzGetOriginFrame(ORIGIN_FRAME_GAME_UI, 0), 0)
```
YakaryBovine, line 25: I'm not sure if we want to take responsibility for the Launcher button away from the Manager. I can easily imagine future Books that are launched without a Launcher button, e.g. the Augment system.

YakaryBovine (commented):
Overall I _think_ your refactor is an improvement, but I'm not sure yet. Will ponder it more. Have left some comments in the mean time.

src/MacroTools/BookSystem/Powers/PowerBook.cs
YakaryBovine: Sorry, to elaborate since I don't think Github's line highlighting made it clear: I'm referring to line 24.

TracyTheBoy: Oh I see. I thought you were referring to the constructor having too many parameters and that being what makes this hard to read.

YakaryBovine: Nah, the many parameters is fine. I was just worried about the visibility being determined via a boolean calculation within the constructor like this, and it not being clear to future Book implementers that that's how you achieve a per-player Book. As it stands I think giving that responsibility to the Book Manager is better separation of concerns.

YakaryBovine:
Closing as #416 supersedes this.

Issue #298, opened 2022-11-14
Make a local version of appsettings.json that overrides the default one if a value is specified


YakaryBovine:
Right now we use the file appsettings.json to configure the launcher.
I'd like it so that the first time we run the project, we get another file called appsettings{USERNAME}.json. This file should be ignored by .git, and any settings in it should override those in appsettings.json.

PR #416, opened 2022-11-20
Book related issues

### LINKED ISSUE #179: Custom frames: Books should reshuffle their contents whenever an item is removed from them

### LINKED ISSUE #281: Organize Artifact menu items by alphabetical order

### LINKED ISSUE #417: Books throw ArgumentOutOfRangeException when spam clicking Next and Previous Buttons
This does not cause any crashes. 
It does print the exception on the screen though.

TracyTheBoy:
Closes #179 
Closes #281 
Closes #417 

Solving #179 #281 and bug that was reintroduced in the commit for fixing #372

I was so unhappy with what I did in the other PR that I decided to just start over.
So this time all I basically did is build on 'IBook' interface.
I made all the buttons that were private before, public and part of the api and I also added methods for each of the buttons expected funtions.
This now allows other systems to work with the book without the use of its buttons.
Its also keeping it flexible by allowing consumer code to decide what is supposed to happen when a button is pressed because they can override the 'OnClick' action of each button.
Right now the `LauncherButton `is part of `IBook` but when that is not needed anymore when the `AugmentBook` is introduced you could easily subclass `Book` or make another interface and move `LauncherButton` along with  `Title` and `LauncherParent` in there and you're good. Or so I think anyway.

All the LauncherButton code in the `ArtifactManager` could be moved to `Book` theoretically but I decided to leave it for now to not screw up the desync issues you fixed in #372.


YakaryBovine:
Brilliant, this is way better. I'll take a proper look after work but I can't see any issues.
P.S. since we've exposing so much Book logic to clients I wonder if we should just take make Book<T> concrete, making it also take in some kind of observable manager as another type constructor. E.g. Book<PageType, ObservableManagerType>. That way Book could be a non-inheritable concrete class that you just configure when you construct it. But, these changes are good even if we don't do that.

src/MacroTools/BookSystem/IBook.cs
```diff
@@ -15,15 +16,54 @@ public interface IBook
...
+
+    /// <summary>
+    /// A button for launching of the book
+    /// </summary>
+    public Button LauncherButton { get; set; }
+
+    /// <summary>
+    /// A button for moving to the next page of the book
+    /// </summary>
+    public Button MoveNextButton { get; set; }
```
YakaryBovine, line 38: I don't think we want to expose the setter on these; I can't imagine a scenario where it's valid for a consumer to arbitrarily modify the Next Button, even if it might be valid to modify what it does. Maybe we should just expose the Get?

LauncherButton seems valid to expose the Set.

TracyTheBoy:
> Brilliant, this is way better. I'll take a proper look after work but I can't see any issues. P.S. since we've exposing so much Book logic to clients I wonder if we should just take make Book concrete, making it also take in some kind of observable manager as another type constructor. E.g. Book<PageType, ObservableManagerType>. That way Book could be a non-inheritable concrete class that you just configure when you construct it. But, these changes are good even if we don't do that.

Yea I like that idea. Having an abstract book class is sort of redundant now that `IBook` exists. 
I did play around with making Book concrete but I ultimately decided to not make too many changes because I wanted to focus on the solving the issues.
And I thought an extra ticket would make more sense for the revision of the system.

YakaryBovine:
Good stuff! P.S. I have the #298 ticket for that appsettings issue ;)

PR #504, opened 2022-11-24
Stonemaul fix

### LINKED ISSUE #481: Stonemaul base instantly turned hostile after Frostwolf completed The Chieftain's Challenge
image
image

This is because the quest goal conflicts with that of Ruins of Stonemaul for Dragonmaw.
image
image

SavanticIO:
Added ObjectiveUnitIsDead
Updated the kill Kor'Gall quests to use new ObjectiveUnitIsDead

Closes #481 

An alternate way to do this would have been to make Kor'Gall a legend and use ObjectiveLegendDead

### DIFF
--- a/src/MacroTools/QuestSystem/UtilityStructs/ObjectiveUnitIsDead.cs
+++ b/src/MacroTools/QuestSystem/UtilityStructs/ObjectiveUnitIsDead.cs
@@ -6,13 +6,13 @@ using static War3Api.Common;
 
 namespace MacroTools.QuestSystem.UtilityStructs
 {
-  public class ObjectiveUnitIsDead : Objective
+  public sealed class ObjectiveUnitIsDead : Objective
   {
     public ObjectiveUnitIsDead(unit unitToKill)
     {
-      trigger trig = CreateTrigger();
-      TriggerRegisterUnitEvent(trig, unitToKill, EVENT_UNIT_DEATH);
-      TriggerAddAction(trig, OnUnitDeath);
+      CreateTrigger()
+        .RegisterUnitEvent(unitToKill, EVENT_UNIT_DEATH)
+        .AddAction(() => { Progress = QuestProgress.Complete; });
       Target = unitToKill;
       TargetWidget = Target;
       InitializeDescription();
@@ -23,12 +23,7 @@ namespace MacroTools.QuestSystem.UtilityStructs
     public override Point Position => new(GetUnitX(Target), GetUnitY(Target));
 
     public unit Target { get; }
-
-    private void OnUnitDeath()
-    {
-      Progress = QuestProgress.Complete;
-    }
-
+    
     private void InitializeDescription()
     {
       if (IsUnitType(Target, UNIT_TYPE_STRUCTURE) || IsUnitType(Target, UNIT_TYPE_ANCIENT))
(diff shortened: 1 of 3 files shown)


src/MacroTools/QuestSystem/UtilityStructs/ObjectiveUnitIsDead.cs
```diff
@@ -0,0 +1,43 @@
...
+using WCSharp.Shared.Data;
+using static War3Api.Common;
+
+namespace MacroTools.QuestSystem.UtilityStructs
+{
+  public class ObjectiveUnitIsDead : Objective
+  {
+    public ObjectiveUnitIsDead(unit unitToKill)
+    {
+      trigger trig = CreateTrigger();
```
YakaryBovine, line 13: Up to you, but you can use Linq-style syntax if you want. E.g:

```cs
CreateTrigger()
   .RegisterUnitEvent(unitToKill, EVENT_UNIT_DEATH)
   .AddAction(trig, OnUnitDeath)
```

src/MacroTools/QuestSystem/UtilityStructs/ObjectiveUnitIsDead.cs
```diff
@@ -0,0 +1,43 @@
...
+                         GetOwningPlayer(Target) == Player(PLAYER_NEUTRAL_AGGRESSIVE);
+    }
+
+    public override Point Position => new(GetUnitX(Target), GetUnitY(Target));
+
+    public unit Target { get; }
+
+    private void OnUnitDeath()
+    {
+      Progress = QuestProgress.Complete;
```
YakaryBovine, line 29: This is so short it might as well just be part of the trigger declaration. E.g.:

```cs
TriggerAddAction(() => { Progress = QuestProgress.Complete; });
```

src/MacroTools/QuestSystem/UtilityStructs/ObjectiveUnitIsDead.cs
```diff
@@ -0,0 +1,43 @@
+﻿using MacroTools.Extensions;
+using MacroTools.FactionSystem;
+using WCSharp.Events;
+using WCSharp.Shared.Data;
+using static War3Api.Common;
+
+namespace MacroTools.QuestSystem.UtilityStructs
+{
+  public class ObjectiveUnitIsDead : Objective
```
YakaryBovine, line 9: Add `sealed` keyword IMO; not designed to be inherited.

YakaryBovine (dismissed):
Good solution; better than using `ObjectiveLegendIsDead` I think. Have left a few syntax suggestions. (I know that code you referenced to create this may not have followed these suggestions)

src/MacroTools/QuestSystem/UtilityStructs/ObjectiveUnitIsDead.cs
SavanticIO: I'm not as used to using Linq-style syntax but makes sense and one to keep in mind.

src/MacroTools/QuestSystem/UtilityStructs/ObjectiveUnitIsDead.cs
SavanticIO: Done

src/MacroTools/QuestSystem/UtilityStructs/ObjectiveUnitIsDead.cs
SavanticIO: Done

YakaryBovine (approved):

PR #513, opened 2022-11-24
13 god off command doesnt actually turn god mode off

### LINKED ISSUE #507: -god off command doesn't actually turn god mode off

SavanticIO:
Closes #507


YakaryBovine (approved):
Nice. Thanks for using the new Cheat base class. I'm surprised that switch expression transpiles correctly!

Issue #461, opened 2022-11-22
Hellfire Citadel should start neutral passive and invulnerable


YakaryBovine:

TracyTheBoy:
I'll take this

Issue #457, opened 2022-11-22
Outland Portals don't teleport you from Eastern Kingdoms to Outland


YakaryBovine:
And possibly other issues.

SebastianTanoni:
Portal should open at turn 10

YakaryBovine:
Looks like this quest from the live version just flat out isn't implemented: https://github.com/AzerothWarsLR/AzerothWarsLR/blob/master/jass/Quests/Fel%20Horde/QuestDarkPortalOpen.j

SebastianTanoni:
Ah yes! there it is, i vaguely remembered i added that quest :)

TracyTheBoy:
I'll take this

Issue #456, opened 2022-11-22
Undercity Portals appear when Lordaeron completes Capital Palace quest


YakaryBovine:

TracyTheBoy:
I'll take this

PR #527, opened 2022-11-24
issued #456 #457 #461

TracyTheBoy:
added `owningPlayer` and `filter` to `RectangleExtensions`. `owningPlayer` allows rescuing units other than neutral passive
 `filter` allows an arbitrary `Func<unit,bool>`  to be passed in. Each unit found in the rectangle is evaluated.  When `filter` returns true the unit is rescued and it is not when `filter` returns false.
I made both parameters optional mainly because I didn't want to change every single call to to this method.
I've assigned default values to both of them so that it still works the same for all existing calls as before

### DIFF
--- a/src/MacroTools/Extensions/RectangleExtensions.cs
+++ b/src/MacroTools/Extensions/RectangleExtensions.cs
@@ -29,33 +29,32 @@ namespace MacroTools.Extensions
     /// </summary>
     /// <param name="rectangle">The rectangle in which to prepare units.</param>
     /// <param name="rescuePreparationMode">Determines how units are prepared.</param>
-    /// <param name="owningPlayer">The player owning the units inside <paramref name="rectangle"/></param>
     /// <param name="filter">A function that is applied to each unit found in <paramref name="rectangle"/>
     /// <para/>
     /// The unit gets rescued if <paramref name="filter"/> returns true, else it does not get rescued.
     /// </param>
     /// <returns>Returns all <paramref name="owningPlayer"/> units in the specified rectangle. Default is neutral passive.</returns>
-    public static List<unit> PrepareUnitsForRescue(this Rectangle rectangle, RescuePreparationMode rescuePreparationMode, player? owningPlayer = null, Func<unit, bool>? filter = null)
+    public static List<unit> PrepareUnitsForRescue(this Rectangle rectangle, RescuePreparationMode rescuePreparationMode, Func<unit, bool>? filter = null)
     {
-      owningPlayer ??= Player(PLAYER_NEUTRAL_PASSIVE);
+     
       filter ??= unit => { return true; }; 
       switch (rescuePreparationMode)
       {
         case RescuePreparationMode.None:
         {
-          return PrepareUnitsForRescue(rectangle, false, false, false, owningPlayer, filter);
+          return PrepareUnitsForRescue(rectangle, false, false, false, filter);
         }
         case RescuePreparationMode.Invulnerable:
         {
-          return PrepareUnitsForRescue(rectangle,true, false, false, owningPlayer, filter);
+          return PrepareUnitsForRescue(rectangle,true, false, false,  filter);
         }
         case RescuePreparationMode.HideNonStructures:
         {
-          return PrepareUnitsForRescue(rectangle,true,true, false, owningPlayer, filter);
+          return PrepareUnitsForRescue(rectangle,true,true, false,  filter);
         }
         case RescuePreparationMode.HideAll:
         {
-          return PrepareUnitsForRescue(rectangle, true, true, true, owningPlayer, filter);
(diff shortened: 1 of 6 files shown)


src/MacroTools/Extensions/RectangleExtensions.cs
```diff
@@ -22,53 +22,62 @@ public static void AddSound(this Rectangle region, sound soundHandle)
...
     /// <param name="rescuePreparationMode">Determines how units are prepared.</param>
-    /// <returns>All Neutral Passive units in the specified rectangle.</returns>
-    public static List<unit> PrepareUnitsForRescue(this Rectangle rectangle, RescuePreparationMode rescuePreparationMode)
+    /// <param name="owningPlayer">The player owning the units inside <paramref name="rectangle"/></param>
+    /// <param name="filter">A function that is applied to each unit found in <paramref name="rectangle"/>
+    /// <para/>
+    /// The unit gets rescued if <paramref name="filter"/> returns true, else it does not get rescued.
+    /// </param>
+    /// <returns>Returns all <paramref name="owningPlayer"/> units in the specified rectangle. Default is neutral passive.</returns>
+    public static List<unit> PrepareUnitsForRescue(this Rectangle rectangle, RescuePreparationMode rescuePreparationMode, player? owningPlayer = null, Func<unit, bool>? filter = null)
```
YakaryBovine, line 38: Sorry, this might be obvious but - could you clarify what problem the new `owningPlayer` parameter is meant to solve? Do we currently have a Quest which needs this?

src/WarcraftLegacies.Source/Quests/Lordaeron/QuestCapitalCity.cs
```diff
@@ -14,7 +15,7 @@ namespace WarcraftLegacies.Source.Quests.Lordaeron
   /// </summary>
   public sealed class QuestCapitalCity : QuestData
   {
-    private readonly List<unit> _rescueUnits;
+    private readonly List<unit> _rescueUnits = new();
```
YakaryBovine, line 18: We don't need the `new()` here since it's being set in the constructor.

src/WarcraftLegacies.Source/Quests/Lordaeron/QuestCapitalCity.cs
```diff
@@ -56,11 +58,12 @@ protected override void OnFail(Faction completingFaction)
     /// <inheritdoc/>
     protected override void OnComplete(Faction completingFaction)
     {
-      completingFaction.Player?.RescueGroup(_rescueUnits);
+      if (completingFaction.Player != null)
+        completingFaction.Player.RescueGroup(_rescueUnits);
       SetUnitInvulnerable(_unitToMakeInvulnerable, true);
       if (GetLocalPlayer() == completingFaction.Player)
         PlayThematicMusic("war3mapImported\\CapitalCity.mp3");
-      LegendLordaeron.Uther?.AddUnitDependency(LegendLordaeron.CapitalPalace.Unit);
```
YakaryBovine, line 63: I think we probably want to keep the null check here, in case this quest is accidentally set up before Lordaeron is.
That said we should find a way to make it impossible for Uther to be null here. Maybe dependency injection?

src/WarcraftLegacies.Source/Quests/Fel Horde/QuestHellfireCitadel.cs
```diff
@@ -4,18 +4,25 @@
...
 {
+  /// <summary>
+  /// Drive the Draenai out of Outland, take control of various control points in outland and upgrade main to t3 in order to unlock Hellfire Citadel.
+  /// </summary>
   public sealed class QuestHellfireCitadel : QuestData
   {
     private readonly List<unit> _demonGates;
     private readonly List<unit> _rescueUnits = new();
 
-    //Todo: mediocre flavour
```
YakaryBovine, line 18: Let's leave these todo comment in. At some point I want to go and fix up these quests with overly generic flavour.

YakaryBovine (changes requested):
Nice work!

src/MacroTools/Extensions/RectangleExtensions.cs
TracyTheBoy: Yes, `QuestHellFireCitadel` requires it because all units start hostile.
We had a similar case when I implemented the "Scepter of The Queen" Quests, but that got fixed in a different way but I can't remember how.

YakaryBovine: Ok, makes sense. `QuestHellfireCitadel` starting hostile is actually a bug - so I think the cleaner move here is to just fix that bug and remove the player parameter from this one. We should be able to guarantee that *all* rescuable units start neutral passive.

Issue #498, opened 2022-11-23
Zandalar and Kultiras second chances are not activating properly


SebastianTanoni:
They do not lose their former holdings

TracyTheBoy:
Kul'Tiras worked properly for me when I killed Boralus as Kul'Tiras

SebastianTanoni:
> Kul'Tiras worked properly for me when I killed Boralus as Kul'Tiras

That means: you lost control of everything you owned prior, you gained the ship moving towards south shore, your heroes revived into the ship and you gained control of all the units inside the ship?

TracyTheBoy:
Yes that's exactly what happened.

TracyTheBoy:
Working on fixing the Zandalar side of this

PR #535, opened 2022-11-26
solved #498

TracyTheBoy:
Extended RectangleExtensions to allow rescuing units of any player. This is needed so that if `QuestConquerKul` is failed and you spawn in ZulFarrak, you don't have to fight all the enemies there.
Refactored other Zandalar Quests to be more in line with the new style and updated them to be in line with the vJass version.

### DIFF
--- a/src/MacroTools/Extensions/RectangleExtensions.cs
+++ b/src/MacroTools/Extensions/RectangleExtensions.cs
@@ -24,7 +24,7 @@ namespace MacroTools.Extensions
     }
 
     /// <summary>
-    /// Prepares neutral passive units within the specified <paramref name="rectangle"/> according to
+    /// Prepares units belonging to <paramref name="owningPlayer"/> within the specified <paramref name="rectangle"/> according to
     /// the provided <see cref="RescuePreparationMode"/>.
     /// </summary>
     /// <param name="rectangle">The rectangle in which to prepare units.</param>
@@ -32,16 +32,18 @@ namespace MacroTools.Extensions
     /// <param name="filter">A function that is applied to each unit found in <paramref name="rectangle"/>.
     /// The unit gets rescued if and only if <paramref name="filter"/> returns true.
     /// </param>
-    /// <returns>Returns all neutral passive units not matching the <paramref name="filter"/> in the specified rectangle.</returns>
-    public static List<unit> PrepareUnitsForRescue(this Rectangle rectangle, RescuePreparationMode rescuePreparationMode, Func<unit, bool>? filter = null)
+    /// <param name="owningPlayer">Player the units inside the specified rectangle belong to. The default value is neutral passive.</param>
+    /// <returns>Returns all units belonging to <paramref name="owningPlayer"/> and not matching the <paramref name="filter"/> in the specified rectangle.</returns>
+    public static List<unit> PrepareUnitsForRescue(this Rectangle rectangle, RescuePreparationMode rescuePreparationMode, Func<unit, bool>? filter = null, player? owningPlayer = null)
     {
       filter ??= _ => true;
+      owningPlayer ??= Player(PLAYER_NEUTRAL_PASSIVE);
       return rescuePreparationMode switch
       {
-        RescuePreparationMode.None => PrepareUnitsForRescue(rectangle, false, false, false, filter),
-        RescuePreparationMode.Invulnerable => PrepareUnitsForRescue(rectangle, true, false, false, filter),
-        RescuePreparationMode.HideNonStructures => PrepareUnitsForRescue(rectangle, true, true, false, filter),
-        RescuePreparationMode.HideAll => PrepareUnitsForRescue(rectangle, true, true, true, filter),
+        RescuePreparationMode.None => PrepareUnitsForRescue(rectangle, false, false, false, filter, owningPlayer),
+        RescuePreparationMode.Invulnerable => PrepareUnitsForRescue(rectangle, true, false, false, filter, owningPlayer),
+        RescuePreparationMode.HideNonStructures => PrepareUnitsForRescue(rectangle, true, true, false, filter, owningPlayer),
+        RescuePreparationMode.HideAll => PrepareUnitsForRescue(rectangle, true, true, true, filter, owningPlayer),
         _ => throw new ArgumentException($"{nameof(rescuePreparationMode)} is not implemented for this function.",
           nameof(rescuePreparationMode))
       };
@@ -56,12 +58,12 @@ namespace MacroTools.Extensions
     /// <param name="hideStructures">If true, prepared structures are hidden.</param>
     /// <param name="filter"></param>
     /// <returns>Returns all neutral passive units not matching the <paramref name="filter"/> in the specified rectangle.</returns>
-    private static List<unit> PrepareUnitsForRescue(this Rectangle rectangle, bool makeInvulnerable, bool hideUnits, bool hideStructures, Func<unit, bool> filter)
+    private static List<unit> PrepareUnitsForRescue(this Rectangle rectangle, bool makeInvulnerable, bool hideUnits, bool hideStructures, Func<unit, bool> filter, player owningPlayer)
     {
       var group = new GroupWrapper()
         .EnumUnitsInRect(rectangle)
         .EmptyToList()
-        .Where(x => x.OwningPlayer() == Player(PLAYER_NEUTRAL_PASSIVE) && filter.Invoke(x))
+        .Where(x => x.OwningPlayer() == owningPlayer && filter.Invoke(x))
         .ToList();
       foreach (var unit in group)
       {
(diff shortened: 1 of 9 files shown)


src/WarcraftLegacies.Source/Quests/Zandalar/QuestGundrak.cs
```diff
@@ -1,39 +1,63 @@
...
+  /// <summary>
+  /// Capture <see cref="LegendNeutral.Gundrak"/> to unlock a new unit
+  /// </summary>
   public sealed class QuestGundrak : QuestData
   {
-    private static readonly int GundrakResearch = FourCC("R02Q");
-    private static readonly int WarlordId = FourCC("nftk");
-    private static readonly int TrollShrineId = FourCC("o04X");
-    
+    private static readonly int GundrakResearch = Constants.UPGRADE_R02Q_QUEST_COMPLETED_THE_DRAKKARI_FORTRESS_WARSONG;
```
YakaryBovine, line 15:
```suggestion
    private static readonly int GundrakResearch = Constants.UPGRADE_R02Q_QUEST_COMPLETED_THE_DRAKKARI_FORTRESS_WARSONG;
```
Might as well do this for other constants as well.

src/WarcraftLegacies.Source/Quests/Zandalar/QuestGundrak.cs
```diff
@@ -1,39 +1,63 @@
...
       "Gundrak has fallen. The Drakkari trolls lend their might to the Zandalari.";
 
+    /// <summary>
+    /// <inheritdoc/>
+    /// </summary
     protected override string RewardDescription =>
       $"300 gold and the ability to train {GetObjectName(WarlordId)}s from the {GetObjectName(TrollShrineId)}.";
 
+    /// <summary>
+    /// <inheritdoc/>
```
YakaryBovine, line 44: Just use a one-line `/// <inheritdoc />` tag and leave out the surrounding `summary` tags.

src/WarcraftLegacies.Source/Quests/Zandalar/QuestZulfarrak.cs
```diff
@@ -1,48 +1,68 @@
...
       AddObjective(new ObjectiveControlLegend(LegendNeutral.Zulfarrak, false));
-      foreach (var unit in new GroupWrapper().EnumUnitsInRect(rescueRect).EmptyToList())
-        if (GetOwningPlayer(unit) == Player(PLAYER_NEUTRAL_PASSIVE))
-        {
-          SetUnitInvulnerable(unit, true);
-          _rescueUnits.Add(unit);
-        }
+      AddObjective(new ObjectiveLegendReachRect(LegendTroll.LEGEND_PRIEST, rescueRect, "Zul'Farrak"));
+      _rescueUnits = rescueRect.PrepareUnitsForRescue(RescuePreparationMode.HideNonStructures);
+      _killUnits = rescueRect.PrepareUnitsForRescue(RescuePreparationMode.HideNonStructures, (unit => GetUnitTypeId(unit) != GetUnitTypeId(LegendNeutral.Zulfarrak.Unit)), Player(PLAYER_NEUTRAL_AGGRESSIVE));
```
YakaryBovine, line 33: I think we want `RescuePreparationMode.None` for this line, as the trolls should start hostile and attackable.

src/MacroTools/Extensions/RectangleExtensions.cs
```diff
@@ -24,24 +24,26 @@ public static void AddSound(this Rectangle region, sound soundHandle)
...
     /// <param name="rectangle">The rectangle in which to prepare units.</param>
     /// <param name="rescuePreparationMode">Determines how units are prepared.</param>
     /// <param name="filter">A function that is applied to each unit found in <paramref name="rectangle"/>.
     /// The unit gets rescued if and only if <paramref name="filter"/> returns true.
     /// </param>
-    /// <returns>Returns all neutral passive units not matching the <paramref name="filter"/> in the specified rectangle.</returns>
-    public static List<unit> PrepareUnitsForRescue(this Rectangle rectangle, RescuePreparationMode rescuePreparationMode, Func<unit, bool>? filter = null)
+    /// <param name="owningPlayer">Player the units inside the specified rectangle belong to. The default value is neutral passive.</param>
+    /// <returns>Returns all units belonging to <paramref name="owningPlayer"/> and not matching the <paramref name="filter"/> in the specified rectangle.</returns>
+    public static List<unit> PrepareUnitsForRescue(this Rectangle rectangle, RescuePreparationMode rescuePreparationMode, Func<unit, bool>? filter = null, player? owningPlayer = null)
```
YakaryBovine, line 37: I think instead of adding another parameter, we should change the trolls in the map to start owned by Neutral Passive, then use the normal `PrepareUnitsForRescue` and make them hostile in the quest constructor. I'm just worried about giving such a ubiquitous function a signature that's even remotely complex. Open to argument.

src/WarcraftLegacies.Source/Setup/QuestSetup/TrollQuestSetup.cs
```diff
@@ -8,14 +10,19 @@ public static class TrollQuestSetup
...
-      zandalar.AddQuest(new QuestJinthaAlor());
-      zandalar.AddQuest(new QuestHakkar());
+      if (zandalar != null)
+      {
+        var questZulFarrak = new QuestZulfarrak(Regions.Zulfarrak);
+        var questZandalar = new QuestZandalar(Regions.ZandalarUnlock);
+        zandalar.StartingQuest = zandalar.AddQuest(questZandalar);
+      
+        zandalar.AddQuest(questZulFarrak);
+        zandalar.AddQuest(new QuestConquerKul(Regions.Zulfarrak, questZulFarrak, new List<QuestData> {questZandalar}));
```
YakaryBovine, line 20: You're only passing in one quest for the last parameter. Maybe just change the signature to `QuestData` instead of a collection?

YakaryBovine (changes requested):
Great stuff. Some minor cleanup to do and one thing to discuss.

src/WarcraftLegacies.Source/Quests/Zandalar/QuestGundrak.cs
YakaryBovine: I don't think my suggestion went through correctly here... I intended to suggest that you use `const` syntax.

src/WarcraftLegacies.Source/Setup/QuestSetup/TrollQuestSetup.cs
TracyTheBoy: Ah yea my bad. I set that up cause I had to potentially add QuestGoldenFleet. Will adjust

src/WarcraftLegacies.Source/Quests/Zandalar/QuestZulfarrak.cs
TracyTheBoy: Yep true, missed that. Will adjust

src/MacroTools/Extensions/RectangleExtensions.cs
TracyTheBoy: The downside of your approach for me personally and anyone else who doesn't know how to work with the editor would be that any cases like would require someone else to do editor work just to get cases like these working.
The function does become more complex but the parameter can be safely ignored 99% of the time since it is optional the default value is the same as before and I'd also argue that this is a very natural way of extending the functionality. 
And I don't see these trolls starting as hostile as 'wrong'. This is just a regular creep camp that just happens to be in the way for when a specific quest fails. 
If we were to remove the quest now, everything would work just fine but if we made the camp neutral passive by default we would either have to do additional editor work or move the function call to a different place where it might not fit.

I don't know if I expressed myself well here. Well this is the downside to not being a native english speaker.

YakaryBovine: I agree. I'm especially sold by your point that without this quest, the units should start hostile.

But I think there is a 3rd option that's best. These units aren't really being "prepared" for rescue; they're not being made invulnerable or hidden, nor are they rescued later. They're just being put into a list so they can be iterated across. So I think it's cleanest to not use PrepareUnitsForRescue at all, as in:
`_killUnits = new GroupWrapper().EnumUnitsInRect(rescueRect).Where(x => x.OwningPlayer == Player(PLAYER_NEUTRAL_AGGRESSIVE))`
This would allow us to keep the `PrepareUnitsForRescue` API clean and restrict it to cases of actual rescue.

TracyTheBoy: I see your point. This would pollute the api and allow 'abusing' the function, open it up too much.

YakaryBovine: Yeah, pretty much. I think we should add the player parameter if and only if there are neutral hostile units that need to be made invulnerable and/or hidden at game start _and_ which must be rescued later. Right now we don't satisfy that category.

YakaryBovine (approved):

PR #538, opened 2022-11-26
14 when plague starts the 3 cultist of the damned should die not removed killed

### LINKED ISSUE #489: When Plague starts, the 3 cultist of the damned should die (not removed, killed)

SavanticIO:
Closes #489

Could be worth investigating exactly why new GroupWrapper().EnumUnitsOfType() wasn't getting the group of units but this was the only place it was being used and the unit list already existed in the PreplacedUnitSystem


YakaryBovine (approved):
Perfect solution IMO. PreplacedUnitSystem having that new public method makes sense; almost surprising we didn't have a use for it earlier!

PR #574, opened 2022-11-29
solved #558 #552

### LINKED ISSUE #552: A MILLION MURGHUL SPAWNED FOR FROSTWOLF
image

### LINKED ISSUE #557: Illidan: Cliff Runner and Glaive Warrior Share Hotkey (Q) in Hunter Proving Grounds

### LINKED ISSUE #558: kael forces got given to Dalaran
image

TracyTheBoy:
closes #558 
closes #552
Put QuelThalasQuestSetup before Dalarans
Removed actions from QuestSeaWitch trigger after triggering once and made it so only Thrall can trigger it

### DIFF
--- a/src/WarcraftLegacies.Source/Quests/Frostwolf/QuestSeaWitch.cs
+++ b/src/WarcraftLegacies.Source/Quests/Frostwolf/QuestSeaWitch.cs
@@ -1,4 +1,4 @@
-using System.Collections.Generic;
+﻿using System.Collections.Generic;
 using MacroTools.Extensions;
 using MacroTools.FactionSystem;
 using MacroTools.QuestSystem;
@@ -41,7 +41,7 @@ namespace WarcraftLegacies.Source.Quests.Frostwolf
         .AddAction(() =>
         {
           var triggerUnit = GetTriggerUnit();
-          if (GetOwningPlayer(triggerUnit) != FrostwolfSetup.Frostwolf?.Player) return;
+          if (triggerUnit != LegendFrostwolf.LegendThrall.Unit) return;
           FrostwolfSetup.Frostwolf.Player?.RescueGroup(_rescueDarkspearUnits);
           var murlocSpawnA = new Point(-3426, -3431);
           var murlocSpawnB = new Point(-4595, -2797);
@@ -58,6 +58,7 @@ namespace WarcraftLegacies.Source.Quests.Frostwolf
           {
             murgul.IssueOrder("attack", new Point(-3372, -2261));
           }
+          TriggerClearActions(_rescueTrigger);
         });
     }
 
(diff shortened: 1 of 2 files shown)


src/WarcraftLegacies.Source/Quests/Frostwolf/QuestSeaWitch.cs
```diff
@@ -41,7 +41,7 @@ public sealed class QuestSeaWitch : QuestData
         .AddAction(() =>
         {
           var triggerUnit = GetTriggerUnit();
-          if (GetOwningPlayer(triggerUnit) != FrostwolfSetup.Frostwolf?.Player) return;
+          if (triggerUnit != LegendFrostwolf.LegendThrall.Unit) return;
```
YakaryBovine, line 44: I think the previous condition is correct - seems to reflect what's in the old GUI version.
image

YakaryBovine (changes requested):
I can't say I would've thought of that quest setup reordering solution but it does make sense!

src/WarcraftLegacies.Source/Quests/Frostwolf/QuestSeaWitch.cs
```diff
@@ -58,6 +58,7 @@ public sealed class QuestSeaWitch : QuestData
           {
             murgul.IssueOrder("attack", new Point(-3372, -2261));
           }
+          TriggerClearActions(_rescueTrigger);
```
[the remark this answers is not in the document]
TracyTheBoy: Yea for some reason I only tried `_rescueTrigger.Destroy()` but that didn't exist

src/WarcraftLegacies.Source/Quests/Frostwolf/QuestSeaWitch.cs
TracyTheBoy: I just thought it'd make more sense thematically for the leader of the horde to recruit the units. We can revert if you want

TracyTheBoy:
> I can't say I would've thought of that quest setup reordering solution but it does make sense!

Wierdly enough it does work the other way around in the vJass version.

src/WarcraftLegacies.Source/Quests/Frostwolf/QuestSeaWitch.cs
YakaryBovine: Yeah, I think we should revert and go with the old condition. If we use this new one we'd have to convey it to the player somehow. After all training Thrall in this scenario is not even mandatory.

src/WarcraftLegacies.Source/Quests/Frostwolf/QuestSeaWitch.cs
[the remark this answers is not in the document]
YakaryBovine: Feel free to add it in `TriggerExtensions.cs`. If you don't I will at some point.

Issue #458, opened 2022-11-22
Implement DestructibleHider from the live vJASS version of the map


YakaryBovine:
Note that this ticket is pretty hard.

SavanticIO:
@YakaryBovine I'll take a stab at this one

PR #576, opened 2022-11-29
16 implement destructiblehider from the live vjass version of the map

SavanticIO:
For issue #458

What this boils down to is spliting the entire map into sub-grids or 'tiles' asigning the tile an Id that can be derived from any coordinate from within the tile. Now all destructables on the map can be mapped to a tileId based on their coordinates.

We add desructables to a list that is mapped to their tileId in a dictionary. Desructables are loaded and unloaded based on where the palyer camera is by accessing the  tile dictionary.


YakaryBovine (approved):
Looks awesome, really well done.

PR #579, opened 2022-11-30
18 add hotkeys to open the artifact and power menus

SavanticIO:
Havent included a hotkey to open power menu but we can revist that if an unused key gets decided on


src/MacroTools/BookSystem/ArtifactSystem/ArtifactBook.cs
```diff
@@ -23,6 +24,14 @@ public ArtifactBook() : base(0.75f, 0.37f, 0.015f, 0.02f)
       Title = "Artifacts";
```
YakaryBovine, line 24: Maybe we should rename this to "Artifacts (F7)"?

YakaryBovine (changes requested):
Code looks perfect. Just added one idea for communicating the hotkey.

src/MacroTools/BookSystem/ArtifactSystem/ArtifactBook.cs
SavanticIO: Great call. Done

PR #1114, opened 2023-01-17
Quest Thandol Span (unfinished)

TracyTheBoy:
The unit it for the destroyed bridge still needs to be added 
Also need testing


src/WarcraftLegacies.Source/Quests/QuestThandolSpan.cs
```diff
@@ -0,0 +1,57 @@
...
+    protected override void OnComplete(Faction whichFaction)
+    {
+      DestructableRestoreLife(ThandolSpan, GetDestructableMaxLife(ThandolSpan), true);
+    }
+
+    private void OnDamaged()
+    {
+      if (!(GetEventDamage() + 1 >= GetUnitState(LegendNeutral.ThandolSpanDestroyed.Unit, UNIT_STATE_LIFE))) return;
+      BlzSetEventDamage(0);
+      SetUnitState(LegendNeutral.ThandolSpanDestroyed.Unit, UNIT_STATE_LIFE, GetUnitState(LegendNeutral.ThandolSpanDestroyed.Unit, UNIT_STATE_MAX_LIFE) * 0.05f);
```
YakaryBovine, line 54: You can shorthand this with an extension method. Something like `LegendNeutral.ThandolSpanDestroyed.Unit.SetLifePercent(0.05f)`

src/WarcraftLegacies.Source/Setup/Legends/LegendNeutral.cs
```diff
@@ -26,7 +26,7 @@ public static class LegendNeutral
     public static LegendaryHero Vaelastrasz { get; private set; }
     public static Capital Caerdarrow { get; private set; }
     public static Capital Shaladrassil { get; private set; }
-    
+    public static Capital ThandolSpanDestroyed { get; private set; }
```
YakaryBovine, line 29: I wonder if there is a good reason to use a Capital for this? You only need the destructable aspect right, so it could just be a plain destructable?

YakaryBovine (changes requested):
review speedrun

src/WarcraftLegacies.Source/Quests/QuestThandolSpan.cs
```diff
@@ -0,0 +1,57 @@
...
+  public class QuestThandolSpan : QuestData
+  {
+
+    private destructable ThandolSpan { get; }
+
+    /// <inheritdoc />
+    public QuestThandolSpan(PreplacedUnitSystem preplacedUnitSystem) : base("Thandol Span",
+      "The massive stone bridges connecting the Wetlands of northern Khaz Modan with the Arathi Highlands of southeastern Lordaeron, known as the Thandol Span is one of the greatest engineering projects ever completed by the dwarves." +
+      "The mighty bridges recently suffered heavy damage due to a Dark Iron. Repair them to connect the two areas once again.",
+      BlzGetAbilityIcon(LegendSentinels.BlackrookHold.UnitType))
```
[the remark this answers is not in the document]
TracyTheBoy: oh yea I just copy pasted that from the other quest. I figured I'd leave that to Sebas

src/WarcraftLegacies.Source/Quests/QuestThandolSpan.cs
```diff
@@ -0,0 +1,57 @@
...
+namespace WarcraftLegacies.Source.Quests
+{
+
+  /// <summary>
+  /// Repair the Thandol Span to make it walkabale again
+  /// </summary>
+  public class QuestThandolSpan : QuestData
+  {
+
+    private destructable ThandolSpan { get; }
```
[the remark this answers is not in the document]
YakaryBovine: Right, sorry, it *is* readonly because it's a get-only private property. But I think it's similar if it's just a readonly field instead.

src/WarcraftLegacies.Source/Setup/Legends/LegendNeutral.cs
TracyTheBoy: I need a unit to be able to use objectives. Or can I convert that somehow?

YakaryBovine: I think you should be able to just use the `ObjectiveControlUnit` objective instead, and use the `PreplacedUnitSystem` to pass in the appropriate unit as a parameter. You would then need to make the unit capturable separately, which you can do with the `MakeCapturable` extension method.

YakaryBovine: Sorry again, I failed to read properly. I see that the objective you're using is actually `ObjectiveLegendReachesFullHealth`. I think we should make a new `ObjectiveUnitReachesFullHealth`, but that could be done after this PR is merged if you'd prefer.

TracyTheBoy: Yea my bad, I don't need a capital anymore. I used a different objective at first which required a capital but not anymore so I'll changed that

YakaryBovine (approved):
excellent

PR #1175, opened 2023-01-22
Reworked half of the cheats to use the new command base class

TracyTheBoy:
- Removed the Augment related cheats (they didn't server a purpose anymore since that system is deprecated).
- Added 1 new Cheat that shows all quest names of a specific faction
- Reworked the `CompleteQuestCommand` to now allow completion of other faction's quests than your own (Example usage: -complete Lordaeron Hearthglen)

Note that the build command ist still broken. Even Removing the condition that we talked about in Discord does not help.


YakaryBovine (approved):
excellent

PR #1473, opened 2023-03-18
Tolvir Custom icons added and moved statholme alter

Technopig1992:
Created and added some basic custom icons for Tolvir buildings. also moved Lordearons stratholme alter slightly to avoid hero getting stuck RE bug thread


TracyTheBoy:
**Important**

Do not merge yet. Binary files will cause merge conflicts which would override changes on main. Will have to resolve this with developers

Issue #1321, opened 2023-02-16
Remove all abilites from units reanimated by persistant soul


TracyTheBoy:

SavanticIO:
I'll have a look at this

PR #1639, opened 2023-04-07
Persistent Soul Fix

### LINKED ISSUE #1626: Revenant revive ability no longer works

SavanticIO:
Made RemoveAllAbilities() the rist action that happens to the reanimated unit.

SetTimedLife() adds a buff which was being removed by RemoveAllAbilities() and basically 'dispelling' the summoned unit.

Should fix #1626 aslo #1321 should be closed.


TracyTheBoy (approved):
Looks good!

PR #1716, opened 2023-04-21
Nzoth Units & building faction setup

Technopig1992:
Nzoth faction and unit constants added for faction setup - Heroes & CP Defender tower variant still missing


Technopig1992:
seems like there some conflicts  , i did pull origin from remote beforehand so unsure why. However only changed VS code nothing on editor so merge should be fine?

PR #1993, opened 2023-07-19
Illidari Blockers

Ripzss:
Added more blockers in Maelstrom by Illidari starting location to prevent Illidan from using Portal to black temple to jump over them.
Glaive Maiden Tooltip fixed
Gate to Silvermoon now starts to be owned by Quelthalas as Finishing the quests made it still to be Neutral.


Ripzss:
Not gate to silvermoon i mean the first Gate to get into Quelthalas

PR #2053, opened 2023-08-31
Parametrize and shorten cinematic duration

grolmash:
Hi!

The intro cinematic and subsequent setup events are currently hardcoded with a few constants in dispersed files.

So I propose instead to define the variable `introSeconds` and pass it to the various setup tasks that require it, replacing hard-coded numbers with arithmetic operations so that changing the duration (within reason) still maintains a proper setup.

This would open the possibility to shorten (60 -> 30) the duration of the cinematic, which feels too long, especially in 8v8 when drafting and joining takes time. With a 30 second cinematic, players would have 16 seconds to choose faction (down from 20), and the same to choose mode (10): this removes 25 second of waiting time between faction and mode choice.

Having a variable cinematic duration requires light changes to how the timers are initialized: instead of setting up a timer that ticks every turn duration (60) from the beginning, we set up one timer of `introSeconds` duration (to display "Game starts in:"), at the end of which the turn timer is initialized _and_ one turn is manually triggered.

Tested for durations from 30 to 60 seconds. 
Changing those can be done simply by editing the parameter (see commit "Shorten cinematic intro").

Note: if introSeconds = 60, nothing changes.

### DIFF
--- a/src/MacroTools/GameTime.cs
+++ b/src/MacroTools/GameTime.cs
@@ -19,12 +19,16 @@ namespace MacroTools
     /// </summary>
     private const float TimerDelay = 2;
 
+
     //This must be after the Multiboard is shown or the Multiboard will break
     private static timer? _gameTimer;
     private static timer? _turnTimer;
+    private static timer? _initialTimer;
     private static timerdialog? _turnTimerDialog;
+    private static timerdialog? _initialTimerDialog;
     private static int _turnCount;
     private static float _currentTime;
+    private static float _introSeconds;
 
     /// <summary>
     /// Fired when a turn ends.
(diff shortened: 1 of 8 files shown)


grolmash:
Fixed the timing for when the intro text appears, as it needs to be triggered after factions are picked.

SavanticIO (changes requested):
Hey.

Thanks for contributing.

On the whole good changes. 
Couple things; 
Could you make introSeconds a const and change the value to 60?
Remove those vscode files.

Cheers,

Sav

YakaryBovine:
The logic changes here are good but we're not looking to shorten the intro. Due to my recent changes your branch now has a bunch of conflicts, so I'll close this PR.
Thankyou for your contribution, and if you feel like updating your branch to resolve the conflicts and reset the intro timer to 60, it would be much appreciated.

PR #2058, opened 2023-09-01
Don't remove protectors

SavanticIO:
Don't remove a unit that is a protector when a player leaves the game

Added UnitIsProtector method which checks if the unit is a protector of any capital

Added UnitIsCapitalProtector checks if unit is a protector of a specific capital


YakaryBovine (approved):
Looks good to me.

PR #2081, opened 2023-09-05
Fix tooltip for Tower Resistance ability

YakaryBovine:
* Renamed Tower Resistance to Piercing Resistance (more descriptive)
* Changed race from Night Elf to Creep (to match Control Points)
* Fix data reference in tooltip


Technopig1992 (approved):
Thanks for adding me as a reviewer good to see that side. PR approved.

PR #2082, opened 2023-09-05
Implemented a test that will raise an error when any model in the map is unused, and deleted all unused models

YakaryBovine:


SebastianTanoni (dismissed):
I could not find a single model that was currently in use, seems to be working perfectly

PR #2084, opened 2023-09-06
Don't remove portals

SavanticIO:
Added new unitType category of 'Portal'
Setup units that need to be categorised as 'Portal'
Don't remove units that are category 'Portal' when a player leaves


src/MacroTools/UnitCategory.cs
```diff
@@ -10,6 +10,10 @@ public enum UnitCategory
     /// <summary>
     /// The unit can train boats.
     /// </summary>
-    Shipyard
+    Shipyard,
+    /// <summary>
+    /// This unit is a portal
+    /// </summary>
+    Portal
```
YakaryBovine, line 17: Looking again at `UnitType`, I think the `Meta` property is exactly what we're looking for here; "If true, this unit should never be deleted." Can we use that instead of using `UnitCategory`?

YakaryBovine (changes requested):
Logic is good but I think we should use the `Meta` property on `UnitType`. Maybe also rename it to `NeverDelete`, lol.

PR #2088, opened 2023-09-07
Never Delete Traders

SavanticIO:
Removed 'portal' from UnitCategory since it isn't being used
Set the NeverDelete flag for traders


YakaryBovine (approved):
Perfect

PR #2102, opened 2023-09-09
Implemented -givegold, -givelumber, and -share commands

### LINKED ISSUE #2094: Add a command to share control, and disable the vanilla share control button
Make sure to include a hint as well.

### LINKED ISSUE #2095: Add commands to give gold and lumber, and disable the vanilla transfer buttons
Make sure to include a hint as well.

YakaryBovine:
Closes #2094 
Closes #2095

### DIFF
--- a/src/MacroTools/Commands/GiveLumber.cs
+++ b/src/MacroTools/Commands/GiveLumber.cs
@@ -8,11 +8,13 @@ namespace MacroTools.Commands
   /// <summary>Give gold to another player.</summary>
   public sealed class GiveLumber : Command
   {
+    private readonly string _commandText;
+
     /// <summary>Initializes a new instance of the <see cref="GiveLumber"/> class.</summary>
-    public GiveLumber(string commandText) => CommandText = commandText;
+    public GiveLumber(string commandText) => _commandText = commandText;
 
     /// <inheritdoc />
-    public override string CommandText { get; }
+    public override string CommandText => _commandText;
 
     /// <inheritdoc />
     public override bool Exact => false;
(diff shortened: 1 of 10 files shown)


src/MacroTools/Commands/GiveLumber.cs
```diff
@@ -0,0 +1,60 @@
...
+    public override bool Exact => false;
+
+    /// <inheritdoc />
+    public override int MinimumParameterCount => 2;
+
+    /// <inheritdoc />
+    public override CommandType Type => CommandType.Normal;
+
+    /// <inheritdoc />
+    public override string Description => "Gives gold to another player.";
```
Technopig1992, line 29: change gold to lumber

src/MacroTools/Commands/Share.cs
```diff
@@ -0,0 +1,45 @@
...
+using MacroTools.FactionSystem;
+using static War3Api.Common;
+
+namespace MacroTools.Commands
+{
+  /// <summary>Share control of your units with another player.</summary>
+  public sealed class Share : Command
+  {
+    /// <inheritdoc />
+    public override string CommandText => "share";
```
Technopig1992, line 12: Might be being pedantic here but would capitalise this so it displays "Share: Shared control with x" rather than "share: Shared control with x"

Technopig1992 (changes requested):
All works fine in game. 
Community are fine with -share working across all allies. 
One typo in GiveLumber

src/MacroTools/Commands/GiveLumber.cs
YakaryBovine: Good catch, thanks

src/MacroTools/Commands/Share.cs
YakaryBovine: You won't see me complaining about useful pedantry. However I tried this change out and I think it's better to use the lowercase version, since it's closer to what the command actually is.

PR #2105, opened 2023-09-10
Unit distribution update

SavanticIO:
Instead of just removing a unit during unit distribution Kill() it as well. Doing this will mean that death events still get triggered.

Set all gate types to 'NeverDelete'

Not sure why the change to MacroTools/UnitCategory.cs is being included looks like that change is already in


YakaryBovine (approved):
Looks good but the conflicts need to be resolved before I can merge. Are you able to do that from your end? Could use the Github UI for it.

PR #2106, opened 2023-09-10
Unit distribution update

SavanticIO:
Instead of just removing a unit during unit distribution Kill() it as well. Doing this will mean that death events still get triggered.

Set all gate types to 'NeverDelete'


YakaryBovine (approved):
Perfect

Issue #2107, opened 2023-09-10
Animate Dead Death knights 


Technopig1992:
add animate dead ability 3 x corpses raised - remove black arrow

PR #2121, opened 2023-09-11
Added Animate Dead to Death Knights - TEST PR 

Technopig1992:
Added Animate Dead to Scourge Death Knights
Removed Black Strike from Scourge Death Knights

Closes - #2107

Testing the new compiler system - large number of files changed apparently (61) . only added 1 ability and edited 1 unit.


mapdata/WarcraftLegacies/AbilityData/Skin/1297035329.json
```diff
@@ -8,16 +8,22 @@
...
       "Id": 2037082209,
       "Type": 3,
-      "Value": "E"
+      "Value": "W"
+    },
+    {
+      "Level": 1,
+      "Id": 828536161,
+      "Type": 3,
+      "Value": "Raises 3 dead units in an area for 60 seconds."
```
YakaryBovine, line 26: This is a vanilla ability so you should use data references in the tooltip, like vanilla does. This one would be something like "Raises <A0OM,DataA1> dead units in an area for <A0OM,Dur1> seconds."

mapdata/WarcraftLegacies/AbilityData/Skin/843460673.json
```diff
@@ -0,0 +1,23 @@
...
+  "OldId": 1684095809,
+  "NewId": 843460673,
+  "Unk": [
+    0
+  ],
+  "Modifications": [
+    {
+      "Id": 1718840929,
+      "Type": 3,
+      "Value": "(Twilight DK)"
```
YakaryBovine, line 11: The data references in this ability tooltip are wrong. They currently refer to the ability `ACad`, which is not this one. Also Twilight isn't in the game, so is there any point adding this?

mapdata/WarcraftLegacies/Units/-1024_-1024.json
```diff
@@ -42,7 +42,7 @@
     "Position": {
       "X": -1359.1619,
       "Y": 4802.799,
-      "Z": 306.30804
+      "Z": 306.30807
```
YakaryBovine, line 45: You're correct that these unit position changes shouldn't be here. Looks like they're all just Z-height changes, so maybe the editor does some annoying shuffling around when you save sometimes. I'll look into it. In the meantime, can you just revert all of the changes done to files in the Units folder?

mapdata/WarcraftLegacies/AbilityData/Skin/1297035329.json
```diff
@@ -8,16 +8,22 @@
     {
       "Id": 1718840929,
       "Type": 3,
-      "Value": "(Twilight DK)"
+      "Value": "(Scourge DK)"
```
YakaryBovine, line 11: You need to update the Death Knight tooltip with this new ability. It still says Black Strike.

YakaryBovine, line 11: The Black Strike ability and its associated buff should be deleted since they're no longer used.

YakaryBovine (changes requested):
Looks good, just a bunch of tweaks to do.

mapdata/WarcraftLegacies/AbilityData/Skin/843460673.json
Technopig1992: Was just ability i copy and pasted for the new DK one. Will remove

mapdata/WarcraftLegacies/AbilityData/Skin/1297035329.json
Technopig1992: does that work on non vanilla abilities too? just out of interest

mapdata/WarcraftLegacies/AbilityData/Skin/1297035329.json
Technopig1992: Okay was not sure about this ill start this practice moving forward.

mapdata/WarcraftLegacies/AbilityData/Skin/1297035329.json
YakaryBovine: It'll work on any object that's defined entirely through the object editor. If it has any C# code defining its behaviour, we probably need to write out values manually.

mapdata/WarcraftLegacies/AbilityData/Skin/1297035329.json
YakaryBovine: Quite reasonable not to be sure! In this case I'm sure Black Strike is only used here.

mapdata/WarcraftLegacies/AbilityData/Skin/1297035329.json
Technopig1992: Changed

mapdata/WarcraftLegacies/AbilityData/Skin/1297035329.json
Technopig1992: Changed

Technopig1992: Black Strike is vanilla Black Arrow it seems and am unable to delete (I assume that's the reason).
Both buffs used are vanilla also.

YakaryBovine:
I think we're going to have to start again with this one, as it has an object ID conflict with Madsen's Vulnerability Aura. In the future we can avoid these by manually specifying object IDs when we create objects.

PR #2123, opened 2023-09-12
Reduce the control point level requirement of "Secrets of Uldum" from 15 to 5

YakaryBovine:
Secrets of Uldum is an amusing quest concept, but it seems to rarely get completed due to how far away it is and the reward tending not to be that impressive. This just makes it more accessible.


Technopig1992 (approved):
Tested multiple times different item each time. All good.

PR #2133, opened 2023-09-14
Frostwolf Grunt Hp This is a Test pr to see if it merges with Herrdave map changes properly 

Ripzss:
Increased from 700->800


YakaryBovine:
@Ripzss to be clear, is this a "real" change that should go into the released map?

YakaryBovine (approved):
LGTM

PR #2140, opened 2023-09-14
I DID IT

HerrDave:
(I learned how to create a pull request!)

- Mulgore Changes
- Duskwood river looks less tropical and more gloomy; mechanically identical
- Fixed Elwynn forest bridges so they match


YakaryBovine (approved):
Really well done with sorting out the new compiler process.
The terrain is amazing, as per usual.

PR #2173, opened 2023-09-18
Reduced Defend Magic Resist

vampirika:
Reduced magic damage reduction of Defend (Universal Ability) from 20% to 15%.


YakaryBovine (approved):
Looks good.

PR #2217, opened 2023-09-23
Herr dave/dalaran revamp (Units & Regions moved)

HerrDave:
- Shifted Silverpine Forest left
- Revamped Dalaran City
- Moved regions over


YakaryBovine (approved):
Great stuff

PR #2230, opened 2023-09-23
Update LegendGoblin.cs

SebastianTanoni:


YakaryBovine:
@SebastianTanoni The change is correct of course.
It only makes sense to put this into `main` if we're going to do a 3.7.5 with this fix at some point. It feels like a reasonably safe bet that we won't at this point. Maybe put this one into `dev`?

SebastianTanoni:
> @SebastianTanoni The change is correct of course. It only makes sense to put this into `main` if we're going to do a 3.7.5 with this fix at some point. It feels like a reasonably safe bet that we won't at this point. Maybe put this one into `dev`?

Sure? i guess theres something i dont understand in the process then. Like when would i ever release into main?

YakaryBovine:
> > @SebastianTanoni The change is correct of course. It only makes sense to put this into `main` if we're going to do a 3.7.5 with this fix at some point. It feels like a reasonably safe bet that we won't at this point. Maybe put this one into `dev`?
> 
> Sure? i guess theres something i dont understand in the process then. Like when would i ever release into main?

You would push directly into `main` if you have a critical bug fix that should be released, but you don't want to release anything that's currently in `dev`. For instance, let's say I find out today that Scourge's defeat condition is broken. I would want to fix that and release it, but I wouldn't want to prematurely release all of the gold changes you've made at the same time.

SebastianTanoni:
> > > @SebastianTanoni The change is correct of course. It only makes sense to put this into `main` if we're going to do a 3.7.5 with this fix at some point. It feels like a reasonably safe bet that we won't at this point. Maybe put this one into `dev`?
> > 
> > 
> > Sure? i guess theres something i dont understand in the process then. Like when would i ever release into main?
> 
> You would push directly into `main` if you have a critical bug fix that should be released, but you don't want to release anything that's currently in `dev`. For instance, let's say I find out today that Scourge's defeat condition is broken. I would want to fix that and release it, but I wouldn't want to prematurely release all of the gold changes you've made at the same time. If you think this is a critical bug fix, we can push this into `main` and release soon. But if it's not, no need.

ok, yeah so it can just go into dev, its a minor fix

YakaryBovine (approved):

PR #2238, opened 2023-09-24
Theramore now has a second chance aspect

SebastianTanoni:
fixed hyptohermic breath tooltip


src/WarcraftLegacies.Source/Quests/Dalaran/QuestTheramore.cs
```diff
@@ -62,6 +63,14 @@ protected override void OnComplete(Faction completingFaction)
         unit.SetOwner(Player(PLAYER_NEUTRAL_AGGRESSIVE));
     }
 
+    protected override void OnFail(Faction completingFaction)
```
YakaryBovine, line 66: You have an `OnFail` method so you need a `PenaltyDescription` and a `PenaltyFlavour`, similar to the reward versions.

YakaryBovine (changes requested):
Looks good, just need to add some more text.

src/WarcraftLegacies.Source/Quests/Dalaran/QuestTheramore.cs
```diff
@@ -45,6 +46,14 @@ public sealed class QuestTheramore : QuestData
     protected override string RewardDescription =>
       "Gain control of all units at Theramore and teleport all of your units within Dalaran to Theramore. Dalaran becomes hostile";
 
+    /// <inheritdoc />
+    protected override string PenaltyFlavour =>
+      "The people of Dalaran have to abandon their city. The colony of Theramore is founded in Kalimdor.";
```
YakaryBovine, line 51: This is actually perfectly solid. Can I make it slightly more emotive, like:
```suggestion
      "Dalaran has fallen. Those who managed to survive its destruction travel west, to the distant lands of Kalimdor. They hope that this new world will treat them more kindly than the one they left behind.";
```

YakaryBovine (approved):
Perfect

PR #2250, opened 2023-09-25
Settings Save/Load System

SavanticIO:
Updated WCSharp.Json and WCSharp.Shared so that WCSharp.SaveLoad could be added.

Added a save and load system using WCSharp.SaveLoad for player settings.

Added commands for toggling Dialogue, QuestText and Captions

Added a command to see current settings


src/MacroTools/Commands/Captions.cs
```diff
@@ -0,0 +1,38 @@
+﻿using MacroTools.CommandSystem;
+using MacroTools.Extensions;
+using static War3Api.Common;
+
+namespace MacroTools.Commands
+{
+  /// <summary>
+  /// A <see cref="CommandSystem.Command"/> that sets the player's camera to a specific height.
```
YakaryBovine, line 8: This comment refers to camera stuff.

src/MacroTools/Commands/Captions.cs
```diff
@@ -0,0 +1,38 @@
...
+
+    /// <inheritdoc />
+    public override string Description => "Toggles the display of captions.";
+
+    /// <inheritdoc />
+    public override string Execute(player commandUser, params string[] parameters)
+    {
+      var captions = parameters[0];
+      if (!bool.TryParse(captions, out var captionsBool))
+        return "You must specify a boolean value(true or false) as the first parameter.";
```
YakaryBovine, line 32:
```suggestion
        return "You must specify either true or false as the first parameter.";
```

src/MacroTools/Commands/Dialogue.cs
```diff
@@ -0,0 +1,38 @@
+﻿using MacroTools.CommandSystem;
+using MacroTools.Extensions;
+using static War3Api.Common;
+
+namespace MacroTools.Commands
+{
+  /// <summary>
+  /// A <see cref="CommandSystem.Command"/> that sets the player's camera to a specific height.
```
YakaryBovine, line 8: Not the right comment.

src/MacroTools/Commands/Dialogue.cs
```diff
@@ -0,0 +1,38 @@
...
+  /// </summary>
+  public sealed class Dialouge : Command
```
YakaryBovine, line 10:
```suggestion
  public sealed class Dialogue : Command
```

src/MacroTools/Commands/Captions.cs
```diff
@@ -0,0 +1,38 @@
...
+    public override bool Exact => false;
+  
+    /// <inheritdoc />
+    public override int MinimumParameterCount => 1;
+
+    /// <inheritdoc />
+    public override CommandType Type => CommandType.Normal;
+
+    /// <inheritdoc />
+    public override string Description => "Toggles the display of captions.";
```
YakaryBovine, line 25: Let's keep the language really simple for people.
```suggestion
    public override string Description => "Turns dialogue captions on or off.";
```

src/MacroTools/Commands/Settings.cs
```diff
@@ -0,0 +1,38 @@
...
+    public override bool Exact => false;
+  
+    /// <inheritdoc />
+    public override int MinimumParameterCount => 0;
+
+    /// <inheritdoc />
+    public override CommandType Type => CommandType.Normal;
+
+    /// <inheritdoc />
+    public override string Description => "Shows the current settings for the player.";
```
YakaryBovine, line 22:
```suggestion
    public override string Description => "Shows your current settings.";
```

YakaryBovine (changes requested):
Really, really cool stuff. Nothing to add in terms of the code I think. I just have a handful of suggestions for tooltips. Nice work!

SavanticIO:
Thanks. I’ve tested the loading, saving and a failed load. I haven’t tested the failedhash response. I think you would have to correctly deserialise, change a value and then serialise again but not 100% sure on that.

YakaryBovine (approved):
Awesome. Happy with this!

PR #2274, opened 2023-09-29
Solarian DragonStorm rename and BloodWarder Feedback Tooltip

### LINKED ISSUE #2109: Sunfury Elites
Tooltip values

Ripzss:
Changed Solarian Draongstorm to -> Lightning Mastery 
Gave new projectile animation to match Solarian projectile animation 
BloodWarder Feedback fixed #2109


YakaryBovine:
I think we need to go back to the drawing board with this one. It's already incorrect that we have both Multishot and Dragon Storm with unique names and identical effects; we can't have a third one. I've updated the Wiki here to make ability requirements clearer: https://github.com/AzerothWarsLR/WarcraftLegacies/wiki/Clarity-standards#abilities
One correct fix here would be to give both abilities the same name and icon, but make it more generic, while still letting them have unique projectiles. Another option might be to replace Solarion's Dragon Storm with a new ability.

PR #2279, opened 2023-09-29
Gate armor changes

SebastianTanoni:
Includes:
a new type of armor, gate armor
The health increase of time
small other changes related to gate balance (cost, time)


YakaryBovine (approved):
LGTM

PR #2282, opened 2023-09-30
First Druid gate did not unlock with the base

SebastianTanoni:


mapdata/WarcraftLegacies/Doodads/1024_10240.json
```diff
@@ -295,7 +295,7 @@
       "Y": 10176,
       "Z": 281.5
     },
-    "Rotation": -1.4017224,
+    "Rotation": 4.881463,
```
YakaryBovine, line 298: Revert changes to Doodads and Units folders please (assuming they're not meant to be here).

YakaryBovine (changes requested):
Only trivial changes needed.

PR #2289, opened 2023-09-30
Vampirika/warsong update blade master attack

vampirika:
Adjusted blademaster attack speed to make it feel more fluid


src/WarcraftLegacies.Source/Constants.cs
```diff
@@ -7,6 +7,9 @@
 /// </summary>
 public static class Constants
 {
+	public const int ABILITY_A0SC_UNHOLY_AURA_RED_DEATH_KNIGHT_ARTHAS_CHOGALL = 1093686083;
```
YakaryBovine, line 10: Are you able to revert all of your changes to Constants.cs please? You didn't change any object names so there should be no change to Constants. Let me know if you need help with this.

YakaryBovine (changes requested):
Actual change looks great, just one issue with the Constants.cs.

src/WarcraftLegacies.Source/Constants.cs
vampirika: I think this is a result of merging with latest dev. Should I not have updated to that?

YakaryBovine: No, that was correct. However I imagine that you had a merge conflict with dev on Constants.cs. The correct solution there would have been to retain all changes from `dev`, and none from yours.
But it's entirely reversible. If you're using Visual Studio I suspect you can just right-click --> reverse. If you're using Github Desktop you don't have that facility, so I would instead copy and paste the entire contents of Constants.cs from `dev` into Constants.cs from your branch.

vampirika: Sometimes Git consfuses me so much xd I merged latest dev again as Sebas had made some changes and I went into Dev branch to grab the constants and copy paste everything into there. Hoping that's sorted it

YakaryBovine: Yeah git sucks. No alternative sadly. This looks good

PR #2290, opened 2023-09-30
Fixed Summon garrison for lordaeron and stormwind not available

SebastianTanoni:


YakaryBovine (changes requested):
To avoid having this issue again in the future, could you add `Stormwind.ModAbilityAvailability(Constants.ABILITY_A0GD_SUMMON_GARRISON_STORMWIND, 1);` to Stormwind's config and `Lordaeron.ModAbilityAvailability(Constants.ABILITY_A0GA_SUMMON_GARRISON_LORDAERON, 1);` to Lordaeron's config? This way the ability availability will be attached to the faction instead of the player slot. At that point I don't think you need the Info.json changes either.

src/WarcraftLegacies.Source/Setup/FactionSetup/LordaeronSetup.cs
```diff
@@ -98,10 +98,10 @@ public static void Setup(PreplacedUnitSystem preplacedUnitSystem)
 
       //Todo: these probably should be in some kind of ability library, not here
       Lordaeron.ModAbilityAvailability(Constants.ABILITY_A0N2_GRASPING_VINES_TREANTS, -1);
-      Lordaeron.ModAbilityAvailability(Constants.ABILITY_A0GD_SUMMON_GARRISON_STORMWIND, -1);
```
YakaryBovine, line 101: What's the removal for? Is the intent that Lordaeron should be able to use Stormwind's Summon Garrison if they get control of one of their buildings?

SebastianTanoni: i brain freezed

PR #2293, opened 2023-09-30
Technopig/crippling shout

Technopig1992:


mapdata/WarcraftLegacies/UnitData/Skin/1329803368.json
```diff
@@ -43,7 +43,7 @@
     {
       "Id": 1651864693,
       "Type": 3,
-      "Value": "Frostwolf Animal Tammer.|n|cfff5962dAbilities:|R Charge, War Cadence|n |n|cff99b4d1Can only train 2.|n|cffffcc00Attacks land and air units.|r"
+      "Value": "Frostwolf Animal Tammer.|n|cfff5962dAbilities:|R Crippling Shout, War Cadence|n |n|cff99b4d1Can only train 2.|n|cffffcc00Attacks land and air units.|r"
```
YakaryBovine, line 46: Could we fix the typos while we're at it, apparently I missed them:
```suggestion
      "Value": "Frostwolf animal tamer.|n|cfff5962dAbilities:|R Crippling Shout, War Cadence|n |n|cff99b4d1Can only train 2.|n|cffffcc00Attacks land and air units.|r"
```

YakaryBovine (changes requested):
Everything looks really solid aside from the fact that it applies Howl of Terror on top of Cripple. Please test that your Targets Allowed change worked then let me know.

Technopig1992:
> Everything looks really solid aside from the fact that it applies Howl of Terror on top of Cripple. Please test that your Targets Allowed change worked then let me know.

Tested, will start making it a part of the practice going forward.  all working okay. its removing the AOE on the ability that avoids it

YakaryBovine:
Well done. Not quite there yet. With a range of 0, the ability can still target the caster. So:
image
I would recommend changing Targets Allowed to Bridge or something.

Technopig1992:
hmm didn't notice that ill make it cripple and it should be fine

YakaryBovine (changes requested):
The last change has clearly not been tested and this is not ready for review.

YakaryBovine (approved):
Looking good.

PR #2343, opened 2023-10-03
Sunfury Buildings & Icons

HerrDave:
- Updated all Sunfury buildings except Well of Eternity
- Removed now unused Sunfury assets
- Updated Sunfury production building names to fit new models
- Fixed Windrunner's Spire icon

*Note
For some reason I got a pull pop-up saying I cannot automatically pull, but can still do so manually?
First time I've seen that, not sure why.


mapdata/WarcraftLegacies/Units/-17920_-23552.json
```diff
@@ -0,0 +1,29 @@
+[
```
YakaryBovine, line 1: @HerrDave Looks like this PR adds a bunch of new unit placements on the map. Is that intentional and if so, what are they?

YakaryBovine (changes requested):
Just one question to be answered, otherwise looks good.

YakaryBovine:
Marked this as a Draft because:
* Sebas needs to look at it before I can do a technical review
* The testing units in the bottom left corner need to be removed

SebastianTanoni:
Units removed, made the small changes i want, everything looks good on my end

mapdata/WarcraftLegacies/Units/-17920_-23552.json
SebastianTanoni: i think they are the testing units, i removed them

YakaryBovine:
I'll just need to take another look at this post-changes.

YakaryBovine (approved):
Looking good!

PR #2382, opened 2023-10-12
Cp towers scaling changed

SebastianTanoni:
control point base damage 19->9
Control point max level 20->30


YakaryBovine (approved):
Looks good.

PR #2416, opened 2023-10-12
Removed all upgrade for flying units, Flight added

SebastianTanoni:
All units that are flying that had an upgrade before now require flight., Flight added to all blacksmith of factions with flying units that had a requirement


YakaryBovine (changes requested):
This is a non-trivial balance change so it should be targeted at `dev` rather than `main`. I think it should be as simple as clicking the button to change the target branch, but if you run into any issues let me know.

YakaryBovine:
image
Quel'thalas still has this research.

YakaryBovine:
image
Stormwind's Blacksmith doesn't say Flight.

YakaryBovine:
Forgeworks doesn't say Flight.
image

YakaryBovine:
Storm Hammers can't be researched, not sure what's happened here.
image

YakaryBovine (changes requested):
Found a handful of surface level issues that need looking at. Note that I only looked at Frostwolf, Stormwind, Quel'thalas, Scourge, and Ironforge. Please look for similar issues for the rest. Need to check
- All air units can still be trained
- All air researches are removed
- All researches related to air units can still be trained
- All Blacksmiths mention Flight in their tooltip

SebastianTanoni:
should be fine, just tested all the factions with flying units

YakaryBovine (approved):
Looking good I think!

Issue #2482, opened 2023-10-18
Fail Scourge's "The Culling" quest when Lordaeron completes "Line of Succession"


SebastianTanoni:
This should be done via a proper objective on The Culling, not by failing it externally.

SebastianTanoni:
@SavanticIO

SavanticIO:
I'll take it

Issue #2474, opened 2023-10-17
Dalaran Waygate


Technopig1992:
A bug that occurs when a linked Waygate is destroyed; the replacement Waygate will not link with the existing primary Waygate

SavanticIO:
Hey. I’ll pick this one up

Issue #2472, opened 2023-10-17
Change the objective for the new Legion Alterac quest to "Scourge has completed Plague of Undeath"


YakaryBovine:
Right now the objective is pretty useless as the game doesn't actually tell you how the quest is completed.

SavanticIO:
Hey not sure if you've looked at this yet but my change for #2482 will be able to cover this as well so I can just do it as part of my changes

Issue #2458, opened 2023-10-16
Gate bug 


SebastianTanoni:
Steps to recreate

kill a gate
then rebuild it
then wait 1 min
it will increment from 500
instead of 2500

500 is the health of a destroyed gate

SavanticIO:
Hey. I’ll pick this one up

PR #2488, opened 2023-10-21
Bug Fixes

SavanticIO:
Fix for issues;
#2474
#2458
#2482
#2472


YakaryBovine (approved):
Looks awesome, thankyou. I might do a little refactoring of the Objective solution because I don't think I want all Objectives to know about their QuestData parent... Will see how avoidable that really is though.

PR #2493, opened 2023-10-22
Update MultiTargetRecall.cs 

SavanticIO:
Don't move units that are resistant
Move units when done casting instead of on cast


SebastianTanoni (commented):
Works flawlessly for me!

SebastianTanoni:
@YakaryBovine il letyou check the code part, but it works on the gameplay side for me

YakaryBovine:
The `IsResistant` part is good. `OnStop` is dangerous because it can run even if the spell hasn't actually been successfully cast, and we can fix the spell by just stopping the teleport from applying to the caster instead. I butchered your fix and applied it in https://github.com/AzerothWarsLR/WarcraftLegacies/commit/c89edbf5a2fa3ce67bb86fab7196dfbc956d90b0

PR #2516, opened 2023-10-25
Thorned Vines Ability

Technopig1992:
Removed Healing Wave from Harvest Witch
Added Thorned Vines ability
Added Thorned Vines ability to Harvest Witch


SebastianTanoni (changes requested):
The slow aura is broken, it accelerates enemy units instead of slowing them down

Technopig1992:
> The slow aura is broken, it accelerates enemy units instead of slowing them down

haha oh dear xD ill fix

YakaryBovine (changes requested):
This is VERY close to being perfect and is quite a cool spell.

This ability is passive and must have a passive icon: 
image

I recommend removing the Slow effect from the debuff; the Vines are clear enough by themselves and we should avoid effects if we can for lag reasons. Up to you though.

I note that the healing given by the Vines is incredibly low. Most units have 0.25 hit point regen, so 30% more gives them 0.075 hit point regen. That's 4.5 hit points **per minute**. I'd recommend removing this part of the spell or massively buffing it. Again, your call.

Technopig1992:
> This is VERY close to being perfect and is quite a cool spell.
> 
> 
> 
> This ability is passive and must have a passive icon: 
> 
> image
> 
> 
> 
> I recommend removing the Slow effect from the debuff; the Vines are clear enough by themselves and we should avoid effects if we can for lag reasons. Up to you though.
> 
> 
> 
> I note that the healing given by the Vines is incredibly low. Most units have 0.25 hit point regen, so 30% more gives them 0.075 hit point regen. That's 4.5 hit points **per minute**. I'd recommend removing this part of the spell or massively buffing it. Again, your call.

Many thanks ! Wow that's pretty underwhelming 😅 I think I'll change the regen to a flat increase rather than percentile , or scrap that part altogether. I'll remove the slow debuff  also . 

Ty

YakaryBovine:
> > This is VERY close to being perfect and is quite a cool spell.
> > This ability is passive and must have a passive icon:
> > image
> > I recommend removing the Slow effect from the debuff; the Vines are clear enough by themselves and we should avoid effects if we can for lag reasons. Up to you though.
> > I note that the healing given by the Vines is incredibly low. Most units have 0.25 hit point regen, so 30% more gives them 0.075 hit point regen. That's 4.5 hit points **per minute**. I'd recommend removing this part of the spell or massively buffing it. Again, your call.
> 
> Many thanks ! Wow that's pretty underwhelming 😅 I think I'll change the regen to a flat increase rather than percentile , or scrap that part altogether. I'll remove the slow debuff also .
> 
> Ty

IMO the healing is pretty cool and you should keep it. And to be clear I'm only suggesting to remove the Slow visual effect, not the entire debuff. Sounds good anyway!

YakaryBovine (approved):
Perfectly implemented, well done.

PR #2518, opened 2023-10-26
Added creep level formula

vampirika:
Added creep level formula based on health, attack type and whether it is flying.


vampirika:
@YakaryBovine to review to see if looks correct implementation. Might still need some adjustments and may be accompanied by some health adjustments on some creeps that seemed to be underpowered.

YakaryBovine:
Yup this looks good. I'll get back to you on the current bug in live.

YakaryBovine:
@vampirika A convenient example of a merge conflict we can look at. This one's resolution is clear though:
1. Merge `dev` back into your branch.
2. Where there is a conflict, accept all changes from `dev`.
3. Re-run W3X to MapData to regenerate those files with the correct level and gold bounty.

YakaryBovine:
Let's re-open this one once work on the next version opens. I suspect it's going to be easier to just copy and paste the migration code over to a new branch rather than trying to resolve the current conflicts.

PR #2533, opened 2023-10-27
Add quest gold to early FW

vampirika:
Giving gold to FW early to temporarily compensate for lower creep gold


vampirika:
@YakaryBovine 

Release notes:

Added 175g to Thunderbluff and Stonemaul quests (350g total).

PR #2554, opened 2023-10-29
Legion early siege on Burning Archer + new ability

SebastianTanoni:


YakaryBovine (changes requested):
It's actually perfect, but I can't merge it in because the buff ID for Burning Arrow conflicts with Vanish.

SebastianTanoni:
should be fixed

PR #2568, opened 2023-10-30
Scarlet Crusade copies Lordaeron's researches when completing The Scarlet Crusade

YakaryBovine:


YakaryBovine:
@SebastianTanoni Note that this is very slightly different to what was discussed. This will copy all researches that are shared by both the sending and receiving Faction, even if it's not registered in the shared Faction config. This is easier to manage in code and I think it should fulfil the purpose anyway. Hope that works for you.

SebastianTanoni:
Cool, il test it when I get home!

SebastianTanoni (approved):
tested, worked flawlessly for all researchs

PR #2575, opened 2023-10-31
Add a "build x unique buildings" quest and use it for all Scarlet Crusade rebuilding quests

YakaryBovine:


SebastianTanoni (approved):
Worked perfectly, much better from a user experience to complete all those quests

Issue #2625, opened 2023-11-05
Stop Ghouls that are in the process of harvesting from being teleported by Plague of Undeath


YakaryBovine:
Feedback tells us that having all Ghouls being teleported is annoying because they tend not to be in Control Groups, so they block the player's real army. We should be able to exclude Ghouls that are currently harvesting lumber using the `GetUnitCurrentOrder` native function.

SavanticIO:
I’ll grab this one.

PR #2628, opened 2023-11-06
Update plague quest

SavanticIO:
Added cheat for getting units current orders
Changed scourge plague quest to not move units that have any resources or their current order is "harvest"

having  GetResourceAmount(unit) > 0 || OrderId2String(GetUnitCurrentOrder(unit)) == "harvest" might seem redundant but units that are given right click commands have a "smart" order and it doesn't change to "harvest" until the first order is complete so this would hopefully catch some units that otherwise would be missed. There is still the chance that harvesters given a "smart" order might still be caught before they can actually gather any resources.


SavanticIO:
This is for issue #2625

YakaryBovine (approved):
Looks perfect. Great cheat to have too.

Issue #2624, opened 2023-11-05
Make it clear in the Plague of Undeath quest text that it only moves units from Northrend


YakaryBovine:
"Army" is a bit of a vague term and players seem to assume it only teleports units from Northrend.

SavanticIO:
I’ll take this one. Do we actually want to update the functionality to only do Northrend units if that’s what most players are expecting?

SavanticIO:
The code has the region to take units from as "Regions.Northrend_Ambiance" so are players just exaggerating, is that region bigger than it should be or is there actaully a bug?

YakaryBovine:
> The code has the region to take units from as "Regions.Northrend_Ambiance" so are players just exaggerating, is that region bigger than it should be or is there actaully a bug?

Interesting... OK. So I think everything is working as intended, and we _actually_ just need to make clear in the quest text that it only takes units from Northrend. I'll update the ticket name.

YakaryBovine:
Fixed by SavanticIO.

PR #2631, opened 2023-11-06
Update QuestPlague.cs

SavanticIO:
Made reward description more clear
Sorted a few issues Rider was complaining about
Switched to use non-obsolete version of IssueOrder()

This is for issue #2624


YakaryBovine (approved):
Perfect

PR #2636, opened 2023-11-07
Scarlet Crusade quest fixes

YakaryBovine:
When transitioning from Lordaeron to Scarlet Crusade...
* The player loses all of their units, rather than them going hostile or transferring any hero experience to allies
* The player retains all of their gold and lumber, and additionally gains the normal 300 gold and 500 lumber as before


YakaryBovine:
@SebastianTanoni I think some of this stuff was still up in the air, so adding you as a reviewer to make sure you're happy with it.
You don't necessarily need to read the code or test anything.

YakaryBovine:
Looked through the Discord history on this one and I think this is pretty much what's intended. If not, easy enough to revisit!

Issue #2629, opened 2023-11-06
Add a delay time to Dreadlord's Dark Summoning spell


YakaryBovine:
**Success criteria:** 
When cast, Dark Summoning teleports 12 units in the target area to the caster after a delay.

**Bonus points**
1. The delay is based on the distance from the caster to the target.
2. If the Dreadlord dies mid-transit, kill a percentage of the transiting units.
3. The visual effect is well-telegraphed, so it's clear where the units are about to land. The Diamond of Summoning ground effect could be good for this.

**Decisions to make**
Should the units arrive at the caster after the delay, or at the location the caster cast the spell at?

YakaryBovine:
The cast time can be achieved just in the Object Editor. However Dreadlords have no channel animation (I think), so we'll need to find an alternate way to show that the spell is being cast so it can be interrupted.

YakaryBovine:
Actually it's probably more feasible to have a transit time, like the real Diamond of Summoning, rather than having a cast time. 
Basically: change Dark Summoning to work like vanilla Diamond of Summoning, just not bugged.

YakaryBovine:
Bonus feature: if the Dreadlord dies mid-transit, kill a percentage of the transiting units.

YakaryBovine:
Bonus feature 2: if the units travel in a straight line from Argus, that implies Argus is closer to SA than it is to NA. Which is cosmological nonsense. Some alternate path of travel if they come from Argus would be fun.

YakaryBovine:
@SavanticIO You get first dibs of this if you want it, since you made the original spell being modified. If you're not keen, I'll take it.

SavanticIO:
Yeah I'll have a go at it

YakaryBovine:
> Yeah I'll have a go at it

Awesome. I think the easiest thing is going to be to use the Buff class. The rest of it should be in the ticket description but let me know if you need any more info.

YakaryBovine:
Thinking about it again - if the effect is not attached to the caster, you're better off with a Hazard object. Unfortunately the Hazard system is something I made so it's less reliable than Buff, but it should be OK.

PR #2653, opened 2023-11-11
Dark Summon Update

SavanticIO:
Changed MultiTargetRecall to DelayedMultiTargetRecall
DelayedMultiTargetRecall uses a buff with a dynamic duration based on distance from the caster, clamped between 3 and 30
If the caster dies before the buff runs out then only a percentage of units are transported, based on remaining buff duration, and  the rest are killed.

There doesn't seem to be a good way to convey remaining time/how long for summon. Maybe one to keep in mind for potential solutions.

For issue #2629


src/MacroTools/Buffs/DelayedRecallBuff.cs
```diff
@@ -0,0 +1,65 @@
...
+        unit.Show(false).SetInvulnerable(true);
+      }
+      
+      Effect = AddSpecialEffect(@"Abilities\Spells\Undead\Darksummoning\DarkSummonTarget.mdl", TargetPosition.X, TargetPosition.Y)
+        .SetLifespan(InitialDuration);
+    }
+
+    public override void OnDispose()
+    {
+      var amountToMove = (int)Math.Floor(UnitsToMove.Count * ((InitialDuration - Math.Floor(Duration)) / InitialDuration));
```
YakaryBovine, line 53: Interesting, so it looks like the amounts of units that die is determined based on the remaining distance to travel. Very cool, but I feel like it'd be easier for the player to understand if it was just a flat percentage of loss if the Dreadlord is dead. What do you think?

YakaryBovine (changes requested):
Overall looks excellent. Couple of minor things to fix, and one discussion point.

src/MacroTools/Buffs/DelayedRecallBuff.cs
SavanticIO: Not as cool but since we cant properly convey all the information to the player it is probably the right way to go about it

YakaryBovine: For sure. Your call, I think it's a good feature either way.

YakaryBovine:
Also it looks like you're not handling the tooltip here so I've added a ticket for follow-up (not necessarily for you): https://github.com/AzerothWarsLR/WarcraftLegacies/issues/2656

YakaryBovine (approved):
Looks beautiful.

PR #2655, opened 2023-11-11
Adding theold god roster in the bottom left area

SebastianTanoni:


YakaryBovine:
We shouldn't be pushing work-in-progress to `dev` unless it's the current area of development focus. Piling units in the bottom left seems innocuous but it's a strain on memory usage just the same as any other group of units.

I'm guessing you want to do this so that you can keep track of the current state of the Old God roster and perhaps share it with others. In which case I recommend just renaming your branch to `Lordsebas/AddOldGods` and leaving it there, like we did with Forsaken. When `dev` is ready for it, it can be merged in.

Issue #2534, opened 2023-10-27
Remove gold gains from Frostwolf's early quests


YakaryBovine:
Recent changes to creep gold impacted Frostwolf's early game gold, so gold was added to their early game quests to compensate for this: https://github.com/AzerothWarsLR/WarcraftLegacies/pull/2533/files
This change should be reverted as soon as a better way can be found, as gold rewards in quests are a blunt force tool that we want to avoid.

YakaryBovine:
Bonus points: document how gold income is meant to work, taking into account the following possible sources, and make it consistent in the map.
 - Control Points
 - Creeps
 - Quest rewards
 - Gold mines

Issue #2295, opened 2023-09-30
Grommash Hellscream's spells are permanently disabled if he uses his abilities in a specific sequence


Ripzss:

SebastianTanoni:
Keep in mind its just a theory, im not 100% sure this is the method to reproduce

Ripzss:
To reproduce Windwalk -> Avatar in windwalk -> Whirldwind on alot of enemites seems to have been a reliable way for me to reproduce this

Ripzss:
Changing Ultimate easiset way to Fix this, 
Suggestions from me:
Change to Galen ult (instant full hp heal)

YakaryBovine:
Unassigning Ripzss just in case anyone is able to pick this up. He's provided reproduction steps. We probably do indeed need to just replace at least one of the abilities involved as the bug seems to be a vanilla one.

YakaryBovine:
Madsen's investigation seems to suggest that this can't be fixed without replacing one of the abilities involved. That makes this a Feature ticket.

Issue #2034, opened 2023-08-27
Demon Gates - Fel Horde


Technopig1992:
Remove limit of 1 on upgraded Demon Gates - Currently you can cheese it anyway and after playing a few games vs 4 Terrorguard gates it does not seem like an issue.

Slightly reduce spawn time for upgraded demon gates

YakaryBovine:
As discussed around here (https://discord.com/channels/1161755984759378100/1161755985363353643/1174112508257255464), Sebas will change the system to avoid this bug occurring. As a result it is now a Feature ticket and must go in a major version.

Issue #2669, opened 2023-11-13
Cthun caster 2 model improvement


SebastianTanoni:
https://www.hiveworkshop.com/threads/arachnathid-optimized-scorpion.329358/

YakaryBovine:
I think this one belongs in the Google Drive? 😆

SebastianTanoni:
> I think this one belongs in the Google Drive? 😆

nono this is a ticket, its not a model dump, its a change im going to do!

PR #2675, opened 2023-11-18
Sebas changes for gold patcj

SebastianTanoni:
Grom ult changed to Field of Blood
Removed the Gold from the Long March
Unit max level 20->30
Added 2 t0 demon gates to make the upgrade linear

#2034 #2295 #2534


YakaryBovine:
Unlinking and re-opening this one because it should be done in `main`: https://github.com/AzerothWarsLR/WarcraftLegacies/issues/2663

YakaryBovine (changes requested):
Looks good. Three comments on the Demon Gate changes.
image
1. The tooltip indicating that you can only build 1, which is on a few of these, is now pointless. Should get rid of that.
2. 2 of the starting Gates are upgraded to Fel Hound gates, which then branch. So the bug described in the Issue still exists.
3. Some of the Demon Gates require Navigation, while others require Stronghold. I'm assuming that's a bug but it's not related to this issue so I'll make a new ticket.

YakaryBovine:
Just noticed another one: QuestStonemaul still says it gives gold. QuestZulfarrak and QuestJinthaAlor both actually give gold, but that could be a separate consideration.

SebastianTanoni:
> Just noticed another one: QuestStonemaul still says it gives gold. QuestZulfarrak and QuestJinthaAlor both actually give gold, but that could be a separate consideration.

Yeah zandalar shrine quest will still give gold for now! Fixed stonemaul though

SebastianTanoni:
#2362 #2701 #2669

YakaryBovine (changes requested):
The icon fix looks good. Full recap of current recap of issues:
1. #2701 should be fixed in `main`, so I'll re-open that issue.
2. #2034 is still bugged due to there being 2 Fel Hound starting gates.
3. Need to remove the "Can only build 1 gate" tooltip since it no longer matters. This might change depending on how (2) gets solved.
4. QuestStonemaul still indicates that it gives "plunder" in the flavour. Need to remove that too.

SebastianTanoni:
@YakaryBovine the bug no longer exists, there is no limit for the gates anymore, so no bug

YakaryBovine:
> @YakaryBovine the bug no longer exists, there is no limit for the gates anymore, so no bug And the tooltip was removed

The limit in the tooltip still exists on my end, for Felguard and Pit Fiend gates at least: 
image

YakaryBovine (approved):
Nice, happy with this.

PR #2739, opened 2024-03-15
Reduced the number of units spawned by the Plague, and buffed some units to compensate

vampirika:
Reduced amount of units produced by plague:

- Necromancers 2 -> 0
- Workers 2 -> 1
- Ghoul 5 -> 2
- Crypt Fiend 2 -> 2 (unchanged)
- Abomination 1-> 2
- Increased periodic spawn of zombies 30s -> 60s
- Removed many villagers and houses that spawned zombies across the land

Stat changes for partial compensation:

Zombie:

- Health 450->550
- Attack base 20->24
- Attack speed 2.7->2.6

Crypt Fiend

- Health 750 -> 850
- Attack base 25 -> 30
- Attack speed 1.8 -> 1.7 cd

Ghoul

- Attack base 11->13


YakaryBovine:
@vampirika We discussed completely removing the periodic Zombie spawns and all of the villager mechanics, which I can do. Does that affect what you've done here negatively?

vampirika:
@YakaryBovine Oh okay might have missed that, thought we were just reducing it. Not particularly, although might want to give back a bit of strength on the spawns but we can add adjust that in later when we see what impact no zombies actually has.

PR #2752, opened 2024-03-18
Gold adjustments

vampirika:
Updated gold formula by reducing high end creep values. Increased increment cost for upgrades. Lowered base cost for caster upgrade.
**Additional gold bounties:**
Creeps over level 7 - 5g -> 3g
Creeps over 1500 - 7g -> 5g
Creeps over 2000 - 10g -> 8g
Creeps over 2500 - 15g -> 10g
E.g. total bonus for 2500hp mob 37g -> 26g

**Upgrade changes:**
Attack upgrade gold increment 55 -> 75
Attack upgrade time increment 25s -> 30s
Defense upgrade gold increment 50 -> 65
Defense upgrade time increment 25s -> 30s
Caster upgrade base gold cost 100 -> 75


YakaryBovine:
I think this will have to wait until I can do those migration changes we talked about! There are so many files here that I can't review any of them.

vampirika:
Makes sense - do you know how long that might be? Otherwise might think of doing some other changes for some races that need it sooner rather than later, and maybe put the upgrade changes into another PR.

YakaryBovine:
> Makes sense - do you know how long that might be? Otherwise might think of doing some other changes for some races that need it sooner rather than later, and maybe put the upgrade changes into another PR.

It'll be a while I think, because I'll be putting those changes into `dev` rather than `main`. So yup I think you'll want to do that split.

YakaryBovine:
Ok @vampirika, I made the change: level and gold bounty data is no longer stored in .json files, since it's calculated from your map migration anyway.

As you can see you now have a tonne of conflicts, but that's no big deal. I recommend just rolling back all of the changes in your the UnitData folder, and we can keep going with the rest.

YakaryBovine (changes requested):
Ok @vampirika, I made the change: level and gold bounty data is no longer stored in .json files, since it's calculated from your map migration anyway.

As you can see you now have a tonne of conflicts, but that's no big deal. I recommend just rolling back all of the changes in your the UnitData folder, and we can keep going with the rest.

vampirika:
@YakaryBovine not sure exactly how meant to do it, so just made a new pr with the changes in

YakaryBovine:
Great, as I understand it we've just merged this in via the other branch so I'll close this one.

Issue #2760, opened 2024-03-18
Quel'thalas adjustments


vampirika:
Reduce magic resist and mana burn of spellbreakers.
Lower wood amount for unit specific upgrades.
Buff other unused units in roster (dragon hawk, sorc)

PR #2763, opened 2024-03-19
Quel'thalas balance

vampirika:
Various balance changes for Quel'thalas. #2760 

**Spellbreakers:**

- Reduced magic resistance 60% -> 55%.
- Reduced base armour 4 -> 3.
- Reduced damage sides per die 4 -> 3.
- Reduced range increase from upgrade 300 -> 250.
- Reduced lumber cost of range upgrade 375 -> 300.

**Sylvanas:**

- Increased ultimate ability number of targets, as was comparitvely weak to others (e.g. tess multishot) 2/3/4 -> 3/5/8.


YakaryBovine:
I gather you've merged these changes into #2769 so I'll just close this one!

PR #2765, opened 2024-03-19
Island changes

SebastianTanoni:
-Removed waterways connecting islands
-removed creeps that are not connected to a control point in the maelstrom -removed Crestfall island, as well as the control point there


YakaryBovine (approved):
LGTM

PR #2767, opened 2024-03-19
Removing lumber sebas

SebastianTanoni:
shredders removed
trading post ability changed
lumber harvesting removed
workers descriptions updated to remove trees


src/WarcraftLegacies.Source/Setup/Spells/SharedSpellSetup.cs
```diff
@@ -24,6 +24,11 @@ public static void Setup()
         },
         Effect = @"Abilities\Spells\Undead\ReplenishMana\SpiritTouchTarget.mdl"
       });
+
+      var tradingPost = new ProvidesIncome(Constants.UNIT_H014_TRADING_POST_SEA, 100);
```
YakaryBovine, line 28: I think we're going to have to think through this implementation a bit more. `ProvidesIncome` is for bonus non-CP income, whereas this Trading Post is now operating just like a Control Point which makes it normal income. I think it probably needs to be integrated into the core Control Point system (even if they don't have CP towers).

Could you revert the Trading Post changes here? I'll spin off a new ticket to deal with those.

YakaryBovine (changes requested):
Just one change needed I think, the rest looks great.

src/WarcraftLegacies.Source/Setup/Spells/SharedSpellSetup.cs
SebastianTanoni: ok, how would you want me to revert them, everything?

SebastianTanoni:
I simply removed the provide income! I kept the rest the same, since if you want to make them like cp, you are going to need the buff and the spell like other control points!

src/WarcraftLegacies.Source/Setup/Spells/GoblinSpellSetup.cs
```diff
@@ -56,9 +56,6 @@ public static void Setup()
       {
         Duration = 10
       });
-
-      PassiveAbilityManager.Register(new ProvidesIncome(Constants.UNIT_NBOT_GOBLIN_PRIVATEER_NEUTRAL_GOBLIN, -2));
```
YakaryBovine, line 60: Is this right? Are these units removed or do they not have that gold drain ability anymore?

mapdata/WarcraftLegacies/BuffData/Skin/B06A.json
```diff
@@ -0,0 +1,34 @@
...
+  "Modifications": [
+    {
+      "Id": 1953653094,
+      "Type": 3,
+      "Value": "ReplaceableTextures\\CommandButtons\\BTNChestOfGold.blp"
+    },
+    {
+      "Id": 1885959270,
+      "Type": 3,
+      "Value": "Command Point 100 gold per minute"
```
YakaryBovine, line 16:
```suggestion
      "Value": "Control Point 100 gold per minute"
```

mapdata/WarcraftLegacies/BuffData/Skin/B06A.json
```diff
@@ -0,0 +1,34 @@
...
+    },
+    {
+      "Id": 1700951398,
+      "Type": 3,
+      "Value": "This unit is a 100 gold per minute CP. It increases your income by 100 gold."
+    },
+    {
+      "Id": 1835101798,
+      "Type": 3,
+      "Value": "Command Point 100 gold per minute"
```
YakaryBovine, line 26:
```suggestion
      "Value": "Control Point 100 gold per minute"
```

YakaryBovine:
> I simply removed the provide income! I kept the rest the same, since if you want to make them like cp, you are going to need the buff and the spell like other control points!

Good point, thankyou! Just a couple of comments in there about the wording then, I think "Command" is surely not right.

mapdata/WarcraftLegacies/Units/9728_8704.json
```diff
@@ -10,21 +10,21 @@
     "AbilityData": [],
```
YakaryBovine, line 10: This file has a conflict because you changed this area in #2765 then are also changing it separately here.
So to fix that I suggest:

1. Pull from `dev` into this branch
2. When it prompts you, overwrite this file with the one from `dev`
3. Redo the change you want in this area (I imagine you removed a Shredder from the map or something)

src/WarcraftLegacies.Source/Setup/Spells/SharedSpellSetup.cs
YakaryBovine: Ah sorry I didn't notice this comment, yup what you've done is good.

mapdata/WarcraftLegacies/Units/9728_8704.json
YakaryBovine: I went ahead and did this for you. It's just easier because I caused a tonne of other merge conflicts with some system changes.

src/WarcraftLegacies.Source/Setup/Spells/GoblinSpellSetup.cs
YakaryBovine: Yeah so it looks like these units indeed aren't used anymore. It's a good cleanup but we should probably just delete the entire unit.

YakaryBovine (approved):
Did a little bit of cleanup myself because I made some fundamental object data changes, so I think this can get merged in now.

PR #2793, opened 2024-03-23
Tree Doodads & Stratholme missing buildings

HerrDave:
- Removed unused tree-related destructibles
- Imported 26 tree models
- Created 26 tree doodads (only used 25: one is an unused barrens tree variant)
- Replaced the missing Stratholme buildings and made them Neutral


YakaryBovine (approved):
Amazing, good work!

Issue #2789, opened 2024-03-22
Remove lumber harvesting upgrades


SebastianTanoni:

SebastianTanoni:
to test

YakaryBovine:
Still need to remove Improved Lumber Harvesting from the following faction configurations:
- Gilneas
- Ironforge
- Lordaeron
- Scarlet Crusade
- Stormwind

Forgeworks tooltip still has it (need to check other Lumber Mill/Blacksmith tooltips):
![image](https://github.com/AzerothWarsLR/WarcraftLegacies/assets/8995690/bce3206b-afd4-4c02-a7bf-4b4e9a664270)

Mason tooltip has it (check other workers):
![image](https://github.com/AzerothWarsLR/WarcraftLegacies/assets/8995690/ad2086dc-0dbd-4d6b-aa94-abfb3b494456)

Also should delete the research itself (i.e. reset it to vanilla values):
![image](https://github.com/AzerothWarsLR/WarcraftLegacies/assets/8995690/4596f043-1c45-4c07-aa16-450fb8fbb623)

And the ability:
![image](https://github.com/AzerothWarsLR/WarcraftLegacies/assets/8995690/134bae4b-9582-4ec8-b6f7-a2fe4f163a13)

PR #2808, opened 2024-03-26
Fixing Lumber harvesting remnants

SebastianTanoni:
#2789 

also removed power infusion from the tooltip of all research buildings


YakaryBovine (approved):
Looking goooood

PR #2810, opened 2024-03-26
Cleaning up the lumber

SebastianTanoni:
Removing some units that were unused and had lumber changed Compensating some lumber costs in gold


YakaryBovine:
I think you probably wanted to remove the other Wisp gather ability. This one is in the wrong button position for its hotkey and has the wrong tooltip (should be the same as every other worker):
image

Same with the Return side of it, which even refers to lumber:
image

SebastianTanoni:
The one i removed was the wrong one, the one that only works with entangled gold mines. So i just need to change this buttons position i guess?

YakaryBovine:
Mountain Giant's Nature Vitality applies Scroll of Protection to the caster:
image

YakaryBovine:
> The one i removed was the wrong one, the one that only works with entangled gold mines. So i just need to change this buttons position i guess?

Yup and the tooltips.

YakaryBovine:
Hold on sorry, I posted these on the wrong PR, deleting

PR #2831, opened 2024-04-01
Balance sweep

SebastianTanoni:


mapdata/WarcraftLegacies/AbilityData/Skin/1496527169.json
```diff
@@ -39,7 +39,7 @@
       "Level": 1,
       "Id": 828536161,
       "Type": 3,
-      "Value": "Adds 50 bonus damage to the unit and causes its attacks to heal it for 80% of damage dealt. |nLasts 25 seconds."
+      "Value": "Adds 50 bonus damage to the unit and causes its attacks to heal it for 80% of damage dealt. |nLasts 60 seconds."
```
YakaryBovine, line 42: Can you make this use data fields?
I ask particularly because it looks like you've changed a field from 50 to 35, and I wonder if that's the bonus damage indicated here.

YakaryBovine (changes requested):
Pretty safe stuff so not going to review it in game. Just one small thing to fix.

mapdata/WarcraftLegacies/AbilityData/Skin/1496527169.json
SebastianTanoni: i changed the mana cost! I also didnt make this spelll, but i can go make fields if you need.

YakaryBovine: Yeah, either fix it or raise a ticket to fix it.
It would be nice to only fix things that we made or broke but I think that'd leave us with a lot of immortal bugs

SebastianTanoni: yeah ok, il fix it now :)

Issue #2889, opened 2024-04-14
Increase gilnease gate unlock region for dalaran quest


vampirika:
More user friendly

Issue #2888, opened 2024-04-14
Correct dalaran towers abilities


vampirika:
Dalaran arrow towers have dark summoning - should be removed.
Dalaran base towers missing reveal - should be added.

Issue #2882, opened 2024-04-13
Outland Area 52 Goblin Shop disappears  


Technopig1992:
Noticed in both Great War and Closed Alliance game modes

Issue #2880, opened 2024-04-13
Dalaran Building Placement 


Technopig1992:
Units can get stuck when trained from the Workshop 

![image](https://github.com/AzerothWarsLR/WarcraftLegacies/assets/127978232/664fc826-ce9e-4bb9-9728-4f197ff84f55)

vampirika:
I'll pick this up

PR #2897, opened 2024-04-14
Minor bug fixes and balance

vampirika:
- Moved sentinel alter blocking heroes. #2878 
- Moved dalaran building blocking units. #2880 
- Remove dark summoning from dalaran towers and add reveal to dalaran base towers (violet towers). #2888 
- Bigger region in front of gilneas gate so heroes can unlock it easier. #2889 
- Reduce dranei piercing reduction. 
  - Piercing deals 0% -> 25%.
- Ashtongue mystic now has unit upgrades and backpack. #2877 
- Fixed goblin merchant being removed when sunfury is not picked. #2882


vampirika:
Was unsuccessful in this change: - Change dreadlord dark summoning so that it can't summon boats. Will need to look into more, maybe another PR.

src/WarcraftLegacies.Source/Factions/Illidari.cs
```diff
@@ -45,7 +45,8 @@ public sealed class Illidari : Faction
         "illidan",
         "ill",
         "naga",
-        "nagas"
+        "nagas",
+        "fish"
```
YakaryBovine, line 49: LOL

YakaryBovine:
Not sure if this was already broken but this tooltip is wrong: 
image
It's also weird that the tower loses Magic Sentry when it's upgraded down this path, but not the other one. Is that intended?

vampirika:
> Not sure if this was already broken but this tooltip is wrong: image It's also weird that the tower loses Magic Sentry when it's upgraded down this path, but not the other one. Is that intended?

almost certainly not, as the improved version has it again.

Will add magic sentry to this tower, and update the tooltip to have all correct abilities (assuming that's all that you see that is 'broken' here?)

YakaryBovine:
> > Not sure if this was already broken but this tooltip is wrong: image It's also weird that the tower loses Magic Sentry when it's upgraded down this path, but not the other one. Is that intended?
> 
> almost certainly not, as the improved version has it again.
> 
> Will add magic sentry to this tower, and update the tooltip to have all correct abilities (assuming that's all that you see that is 'broken' here?)

Yep! I think adding Magic Sentry and updating the tooltip covers the bases.

src/MacroTools/Extensions/RectangleExtensions.cs
```diff
@@ -66,6 +66,7 @@ public static void CleanupNeutralPassiveUnits(this Rectangle area, NeutralPassiv
         }
 
         if (cleanupType == NeutralPassiveCleanupType.RemoveUnits || unit.IsType(UNIT_TYPE_STRUCTURE))
+          if (unit.GetTypeId() != FourCC("ngme"))
```
YakaryBovine, line 69: As discussed on Discord, let's make this logic generic rather than relying on a hardcoded FourCC. The other one I put in this function is not great either. 

Instead we can check to see if the unit being iterated over is `NeverDelete` and invulnerable. If that's the case, we don't need to do anything to clean the unit up. We'll also have to make sure Goblin Merchant is registered as `NeverDelete` in `UnitTypeConfig`.

YakaryBovine (changes requested):
Tested everything and it works well. Just a few things to clean up.

YakaryBovine (approved):
Looking good

PR #2924, opened 2024-04-20
Frostwolf tooltip and bug fixes

vampirika:
Added upgrades to serpent wards.
Generalised tooltips to not include damage values that were incorrect.
Removed navigation requirement from animal companion upgrade (trolls) and grasping vines upgrades (druids).


YakaryBovine (approved):
Perfect

PR #2937, opened 2024-04-24
Balance changes

vampirika:
Druids:

- Reduced dryad upgrade costs. 
  - 100g -> 50g
- Buffed dryad poison upgrades. 
  - Crippling poison attack speed slow up to 40% (previously 35%, was 40% after 2nd upgrade)
  - Deadly poison upgrade increased dps 5->6 but reduced duration 5->4; total damage of DoT 20->24.

Warsong:

- Fixed grom stats after bloodpact matching regular form. (_oopsie_!)

Gilneas:

- Reduced tess W attacked speed buff 200/300/400/500% -> 175/250/325/400%.
- Fixed a bug with tess shooting too many projectiles at low levels.

Scourge:

- Increased efficacy of anub'arak ultimate
  - Number of locusts from 20/30/40 -> 25/35/45.
  - Locust attack cooldown down 1.0 -> 0.5.
  - Reduced heal threshold from 20/30/40 to 20 at all levels. (Locusts will return to anub'arak sooner).
  - Locust unleash instantly.
- Increased strength of beetles at higher levels
  - Beetles level 2 health 550 -> 650.
  - Beetles level 3 health 650 -> 850.
  - Beetles level 4 uses level 3 beetles (as before) but now spawns 4 instead of 2.
  - Beetle manacost lowered to 30 at all levels.

Lordaeron:

- Removed West goldmine for lordaeron, increased starting goldmine by 200g. (-400g overall).
- Removed some creeps associated with lordaeron creeping, particularly in scholomance.


YakaryBovine (approved):
Looks clean, well done.
I'll be very interested to hear if the Anub'arak buffs have the intended outcome. Locust Swarm feels way nicer to use now with the swarm coming out instantly.

PR #2951, opened 2024-04-24
Implement Incubate ability

YakaryBovine:
**Incubate:** Raises an Immature Egg from a nearby corpse. After 1 minutes the egg matures. Mature eggs can be hatched into permanent Silithid Soldiers, and will automatically hatch if they are destroyed.
**Rapid Incubation:** Causes Immature Eggs incubated by Vile Corrupters to mature in 1 minute instead of 5.


SebastianTanoni (approved):
Tested everything. spells works fine, upgrades looks fine

The only thing i noticed: The defensive cocoon disbtn looks like a PAS one, so at first i thought the icon was wrong and it was set to passive in the altar, but it was just disabled. Not a big deal but i mention it anyway!

YakaryBovine:
> Tested everything. spells works fine, upgrades looks fine
> 
> The only thing i noticed: The defensive cocoon disbtn looks like a PAS one, so at first i thought the icon was wrong and it was set to passive in the altar, but it was just disabled. Not a big deal but i mention it anyway!

Yeah it's a good point. I'll see if I can find a different egg icon

PR #2957, opened 2024-04-25
Added Spell Conduction and Hardened mutually exclusive researches to Obsidian Eradicators

YakaryBovine:
**Spell Conduction:** Allows Obsidian Eradicators to use Spell Conduction, which causes 35% of Magic and Spell damage dealt to friendly units within a radius of 500 to be redirected to them.

**Shaped Obsidian:** Changes Obsidian Eradicators' armor type to Heavy and replaces their Spell Resistance with Hardened, which causes them to take 30 reduced damage from attacks.


SebastianTanoni (approved):
Tested the research, they are exclusive.

Tested the passive, it correctly redirects damage.

PR #2958, opened 2024-04-25
YoggSaron world boss, quest and item

SebastianTanoni:


YakaryBovine:
Wow, this is really cool, good effort. Tooltip and icon and everything is super clean and the Artifact does a tonne, which is great for a world boss.

Only found one thing to fix, which is that the item says it gives me Mass Haste but it really gives me Rally:
Untitled

YakaryBovine (changes requested):
Just a single fix needed, great stuff.

YakaryBovine:
Need to update the buff to correspond with the new ability name: 
image

PR #2972, opened 2024-04-30
Uldum Terrain

HerrDave:
- Terrained Throne of the Winds
- Terrained Vortex place
- Terrained Uldum
- Terrained Sholazar Basin to prevent catapults from attacking Icecrown from outside
- Removed more destructible trees
- Removed creeps around the Lost City of the Tol'vir
- Created new Doodad: Vortex Tower
- Created new Region: Vortex Rain
- Placed Neutral Passive elemental units in the new terrain


YakaryBovine (approved):
I haven't looked at this in-game yet (I'm excited to!) but from the list of changes I can see that it's pure terrain, which I don't need to review the quality of. So approved!

PR #2985, opened 2024-05-01
Sunfury Altar Fix

HerrDave:
- Extracted, fixed and reimported Sunfury's Altar (BE_Altar.mdx)

I think the Uldum Terrain included itself in this branch as well, but I've unchecked all of the boxes except the ones leading to war3mapImported, Units and Source


YakaryBovine (approved):
Brilliant! And yup, what you did with git makes sense.

Issue #2975, opened 2024-05-01
SW can still research Magic Sentry


Ripzss:
![image](https://github.com/AzerothWarsLR/WarcraftLegacies/assets/105180704/2fede1e3-3434-41d1-ab45-f84255bc2d76)

Issue #2934, opened 2024-04-23
The Blood of Mannoroth ability does not actually tell you what it does!


Technopig1992:
![image](https://github.com/AzerothWarsLR/WarcraftLegacies/assets/127978232/ffbb5f50-d57a-490a-b3b6-4044a091b725)

PR #2986, opened 2024-05-02
Bugs

### LINKED ISSUE #2899: Missing Pathing Blockers
image
image

Not sure with the last one as you are able to walk through the gate?
image

Technopig1992:
Magic Sentry removed from SW
Drink the Blood of Mannaroth tooltip updated
Pathing blockers added 

Closes
#2899 
#2934 
#2975


YakaryBovine:
Need to remove Magic Sentry from this tooltip: 
image

YakaryBovine:
So, this is good, and I don't need you to fix anything, but be careful with quest research tooltips like this:
image

Ultimately the quest itself already has all the information as to what completing it does, the research is just a stepping stone to get there. And if we include the information in both places, we have to keep it up-to-date in both places. That's why for other ones I did something like this:
image

YakaryBovine:
Pathing blockers feel good.

YakaryBovine (changes requested):
Couple fixes needed!

Technopig1992:
Ill let you off , turns out blizzard doesn't automate them xD 

With the quest research I could remove the second paragraph and replace with "Required to complete the Blood Pact quest" if you like? I based it off the Plague research originally.

YakaryBovine (approved):
Nah, I'm happy with this. It's accurate and that's the main concern. Just something amusing to refer back to when it inevitably gets out of sync 🗡️

PR #2999, opened 2024-05-05
KT and Fel Bugfixes

vampirika:
Added flight research for KT.
Removed range from incinerate upgrade.


YakaryBovine (approved):
Looking good!

Issue #3007, opened 2024-05-09
Black Empire model issues


SebastianTanoni:
Fateweaver doesnt have a portrait

Obelisk is invisible and slowly appears,.

PR #3011, opened 2024-05-09
Black empire Unit and buildings, some upgrades

SebastianTanoni:


YakaryBovine:
image
Need to remove all references to lumber! So check Black Empire's Town Halls, workers, and Blacksmith.

YakaryBovine:
image
I presume this should be "Unworldly" rather than "Unwordly"?

YakaryBovine:
I think you probably want their damage type to start off as Hero since they're elites:
image

YakaryBovine:
This cheeky tentacle on the left attacked me: 
image

YakaryBovine:
image
Should just be Far Sight!

YakaryBovine:
Should be Obliterate (mechanically identical): 
image

SebastianTanoni:
> image Need to remove all references to lumber! So check Black Empire's Town Halls, workers, and Blacksmith.

Also missing the F building

YakaryBovine:
image
These Obelisk models are funky, it takes like 10 seconds to show anything!

YakaryBovine:
Costs lumber: 
image

SebastianTanoni:
> image These Obelisk models are funky, it takes like 10 seconds to show anything!

https://github.com/AzerothWarsLR/WarcraftLegacies/issues/3007

YakaryBovine:
image
Shouldn't mention destroying trees

YakaryBovine:
image
This guy disappears as soon as I train him

YakaryBovine:
These two buildings have the same hotkey: 
image

YakaryBovine:
The tooltip here is "Research Accelerated Cycle" but the research name is actually Cursed Flesh.
You don't need to do anything here because I'll just automate it, but this is an FYI
image

YakaryBovine (changes requested):
Looks great, not much to fix

YakaryBovine:
What's "Old Scroll of Teleportation"? I imagine we need to delete that? Other shops have a different version.
image

YakaryBovine:
Need to move these units to Q and W positions: 
image

YakaryBovine:
Button positions here are all kinds of screwy: 
image

YakaryBovine:
Need to capitalize One here: 
image

YakaryBovine:
Looks like the buff tooltip here got cut off. I imagine you were caught out by the fact that WC3 cuts off tooltips after the first comma!
image

YakaryBovine:
These two buildings have the same icon: 
image

YakaryBovine:
Weird hotkey here: 
image
Might automate hotkeys at some point, but not sure if I should automate hotkey from button position or button position from hotkey...

YakaryBovine (changes requested):
Found some more!

YakaryBovine:
Images aren't uploading right now so this is just a comment.
 - Herald's Chaos Siphon says it can raise the caster's mana above its maximum value, but it actually can't. Need to update the tooltip or fix the spell.
 - Remove the first sentence from Chaos Siphon; "transfers Chaotic Energy" implies that Chaotic Energy is a mechanic.
 - I can't figure out how to increase the level of Heal. Need to either add a way to upgrade it or remove the "Level 1" bit from the tooltip.

YakaryBovine (changes requested):
Definitely getting into the final touches now!

YakaryBovine (approved):
OK, looking great!!

PR #3033, opened 2024-05-14
Technopig/bugs

Technopig1992:
- Rematerialization no longer works on units with Resistant Skin as intended.
- Outpost are now awarded on quest fail as intended as for the following:
Warsong, Uldum Outpost
Golbin, Ratchet Outpost
Zandalari, Ratchet Outpost


src/MacroTools/Powers/Rematerialization.cs
```diff
@@ -50,10 +50,12 @@ private void OnUnitDeath()
       if (GetRandomReal(0, 1) > _chance 
           || !EligibilityCondition(dyingUnit) 
           || _noReturnRect.Contains(dyingUnit.GetPosition())
+          || dyingUnit.IsType(UNIT_TYPE_RESISTANT)
```
YakaryBovine, line 53: Let's update the Description accordingly, so like "Your non-Resistant units have a {chance*100}% chance to rematerialize at {returnPointName} upon death.";

YakaryBovine (changes requested):
Looks great functionally, just needs an updated tooltip.

mapdata/WarcraftLegacies/UpgradeData/Skin/909717586.json
```diff
@@ -39,7 +39,7 @@
       "Level": 1,
       "Id": 828536167,
       "Type": 3,
-      "Value": "Grants your demon units a 20% chance to rematerialize at Argus upon death."
+      "Value": "Grants your non-Reseistant units a 20% chance to rematerialize in Argus upon death."
```
YakaryBovine, line 42:
```suggestion
      "Value": "Grants your non-Resistant units a 20% chance to rematerialize in Argus upon death."
```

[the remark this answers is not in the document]
Technopig1992: Ah it's back ? Very cool, I miss that little menu . Kk will do ! Ty

src/WarcraftLegacies.Source/Quests/Dalaran/QuestNewGuardian.cs
```diff
@@ -18,7 +18,7 @@ public sealed class QuestNewGuardian : QuestData
       @"ReplaceableTextures\CommandButtons\BTNAstral Blessing.blp")
     {
       _jaina = jaina;
-      AddObjective(new ObjectiveControlCapital(dalaran, false));
+      AddObjective(new ObjectiveCapitalAlive(dalaran));
```
YakaryBovine, line 21: This works way better but, looking at the list of Objectives, I actually don't think we need a new Objective for it. Use `ObjectiveUnitAlive` instead - it's the same one used in Dalaran's Theramore quest anyway.

YakaryBovine (changes requested):
I've had a chance to test everything now and it works great, just one tidy-up comment to resolve.

src/WarcraftLegacies.Source/Quests/Dalaran/QuestNewGuardian.cs
Technopig1992: Gotcha. Shall  I remove the ObjectiveCapitalAlive file now also as its now not it in use

YakaryBovine: Yup. The reason we have a distinct `ObjectiveLegendAlive` objective rather than just using `ObjectiveUnitAlive` is because Legends can sometimes change units, whereas Capitals always have the same unit. So `ObjectiveCapitalAlive` isn't needed.

YakaryBovine (approved):
Looks great!

PR #3040, opened 2024-05-15
Town halls distance from gold mine

SebastianTanoni:


YakaryBovine (approved):
Happy with the list of files changed, and Sebas has tested this thoroughly in-game so... tick!

PR #3052, opened 2024-05-23
Code point change for portals location

SebastianTanoni:


YakaryBovine:
As discussed these PRs will receive a bit of a tidy-up so just marking as draft.

PR #3053, opened 2024-05-23
Terrain Changes Slip

SebastianTanoni:


YakaryBovine:
As discussed these PRs will receive a bit of a tidy-up so just marking as draft.

YakaryBovine:
Definitely seems to have helped. I can still cause the pathing issue with the Legion teleporter easily enough, but I'm hoping once I force the spawn point for it that will be solved.

Issue #3050, opened 2024-05-22
Slipstream waygate pathing is highly unreliable


YakaryBovine:
From investigating #2243 we've found that Slipstreams with waygate destinations or origins with cluttered pathing sometimes don't allow units through even when it looks like they should. To fix this and make it reliable for players we need to:

1. Ensure that all pre-set Slipstream destinations and origins are cleared of non-pathable doodads and terrain.
2. Change Exodar and Legion Teleporter Slipstreams to have a forced origin point rather than letting the player target it _(this will let us be sure that it's valid)_.

List of Slipstreams are as follows:
- Illidan's innate spell
- Legion Teleporter spells
- Draenei Dimensional Generator spells
- Item gained from Frostwolf's "Jailor of the Damned"

As this is a high complexity bug fix it will be going into the `dev` branch.

YakaryBovine:
I'll do the spell/code portion and Sebas has volunteered to do the terrain changes.

YakaryBovine:
We've cleared out the terrain from the origin and destination locations, and made the origin points of the Slipstreams static, but we're still having trouble with the Argus to Alterac portal. In `dev` right now you can reliably go from Alterac to Argus, but the other direction consistently doesn't work.

YakaryBovine:
Sebas will try randomly moving the Alterac portal around until it works reliably. If that doesn't work, we'll have to brainstorm another idea.

YakaryBovine:
Our attempts to fix this haven't really worked. It's still _very_ easy to create an issue, just...

1. Create either of the Legion portals.
2. Cancel it.
3. Create either of the Legion portals.
4. Attempt to go through; your units will just gather at the portal, then give up and try walk to the destination. I tried this with 12 Footmen.

PR #3055, opened 2024-05-25
Legion Teleporter and Dimensional Generator create their Slipstream portals in a static location

YakaryBovine:


YakaryBovine:
Related to, but does not close, #3050

Issue #3061, opened 2024-06-02
Storm Crow Form should require Flight


YakaryBovine:
![image](https://github.com/AzerothWarsLR/WarcraftLegacies/assets/8995690/9b17b13c-7d0b-4926-a8a8-eace7db6d3d0)

Issue #2869, opened 2024-04-11
Remove extra workers from quests that's not linked to a Gold Mine


Ripzss:
- [x] Frostwolf (Echo isle move to Thunder bluff Gold Mine)
- [x] Druids (base proteted by wildkin)
- [x] Sentinels (auberdine)
- [x] Stormwind (Goldshire, Lakeshire, Darkshire, NetherGarde Keep)
- [x] Scourge (Draktharon move it to the frozen throne Gold Mine)
- [x] Illidan (STV base unlock)

Technopig1992:
With the Illidan STV ones I would either leave them or replace with a Town Hall in STV. Otherwise Illidans only workers come all the way from BT

PR #3062, opened 2024-06-03
Workers, Storm Crow, Katherine Make it Rain

Ripzss:
Deleted/ moved some workers around as some was there for lumber gathering.

Storm Crow upgrade now requires flight

Kathering Make it Rain was based on Rain of Fire and had damage per second on the wave making it alot stronger than other abilities like Blizzard from jaina now it should be more in line with blizzard

Related Issues
#3061 #2869


YakaryBovine (approved):
Looks good!

PR #3071, opened 2024-06-11
Slipstream should now work

SebastianTanoni:


YakaryBovine:
As discussed on Discord, needs to be redone pointing to `dev` or this will blow away the other changes we made to fix #3050

PR #3076, opened 2024-06-17
Vampirika/NA balance

vampirika:
### General

- Reduced Battle for Azeroth timer by 2 minutes (now turn 14 instead of 16 for fog of war)

### Creeps

- Added some additional low level trolls in quel creep camps
- Moved creeps in Zul'aman so they are in smaller groups
- Removed some creeps across hinterlands, arathi areas

### Lordaeron

- Reduced starting level of Garithos and Morgraine 7 -> 6

**Uther:** Reduced overall power

- Base strength decreased 40 -> 30
- Base intelligence decreased 31 -> 28
- Consecration (Q) (Also affects Brigitte Abendis, Scarlet)
  - Cooldown increased 6 -> 7
  - Manacost increased 90 -> 100
  - Damage decreased later 60/120/180/240 -> 70/120/170/220 (tooltip corrected)
- Lightbringer strength (W)
  - Armor increased 4/6/8/10 -> 4/7/10/13
  - Health regenration increased 10/15/20/25 -> 10/16/22/28
  - Mana cost reduced 50/60/70/80 -> 45/50/55/60

**Arthas:** Reduced king arthas power
- Base strength decreased 42 -> 38
- Strength per level decreased 3.5 -> 3.4
- Base intelligence decreased 38 -> 34
- Solar Judgement (R)
  - Duration decreased 14 ->10
  - Damage decreased early 30/50/70 -> 25/50/75

### Quel'thalas

- Reduced anasterian defensive orb (E) damage by 5 at all levels
- Sylvanas base agility increased 21 -> 24
- Sylvanas agility per level increased 2.0 -> 2.5
- Starting goldmine value decreased 1000 -> 800
- Adjusted creeps in quel'thalas (see above)

### Scourge

- Small QoL changes to creeping.
  - Change some golems to wolves in crystalsong forest.
  - Removed nerubian queen from icecrown glacier cp (Queen Nezar'Azret already rules there). Placed mammoths in storm peaks to compensate gold/xp.
- Baron rivendare tankiness increased
  - Base strength increased 23 -> 27
  - Base agility icnrease 15 -> 18

### Legion

- Fel guard fel stalker summon weaker
  - Health reduced 750 -> 650
  - Base damage reduced 23 -> 20
- Archimonde stronger
  - All base stats increase by 10
  - Starting level increased 8 -> 10

### Bug Fixes

- Added "base" to protection aura (stormwind siege tower) tooltip i.e. only armor without upgrades is affected
- Moved troll town hall in echo island to avoid blocking.
- Added mana regen to kul'tiras tier 3.


src/WarcraftLegacies.Source/Setup/Spells/LordaeronSpellSetup.cs
```diff
@@ -25,9 +25,9 @@ public static void Setup()
       
       var solarJudgement = new SolarJudgementSpell(ABILITY_A01F_SOLAR_JUDGEMENT_LORDAERON_ARTHAS)
       {
-        DamageBase = 10,
-        DamageLevel = 20,
-        Duration = 14,
+        DamageBase = 0,
+        DamageLevel = 25,
```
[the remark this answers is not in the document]
YakaryBovine: Discussed on Discord. Conclusion: this is actually a very good change. Lordaeron heroes are meant to scale really well but not be overwhelmingly strong early, and this puts us another step in the right direction!

YakaryBovine (changes requested):
image
We should avoid normalizing mana costs bit-by-bit like this, and instead either do it all at once or not at all. Otherwise we'll end up with a situation where some abilities have scaling mana costs and others do not, and the distinction between them won't be for any reason. And we'll put way more effort into it than we need to (we can just write a migration script!),

Other than that, looks good to me. There's nothing I really feel the need to test since I know you're on the ball about testing tooltips.

src/WarcraftLegacies.Source/Setup/Spells/QuelthalasSpellSetup.cs
```diff
@@ -60,7 +60,7 @@ public static void Setup()
         OrbitRadius = 350,
         OrbitalPeriod = 4,
         OrbEffectPath = @"war3mapImported\OrbFireX.mdx",
-        Damage = new LeveledAbilityField<float> { Base = 25, PerLevel = 25 },
+        Damage = new LeveledAbilityField<float> { Base = 20, PerLevel = 25 },
```
YakaryBovine, line 63: I feel like there should be a corresponding tooltip change for this.

vampirika:
> image We should avoid normalizing mana costs bit-by-bit like this, and instead either do it all at once or not at all. Otherwise we'll end up with a situation where some abilities have scaling mana costs and others do not, and the distinction between them won't be for any reason. And we'll put way more effort into it than we need to (we can just write a migration script!),
> 
> Other than that, looks good to me. There's nothing I really feel the need to test since I know you're on the ball about testing tooltips.

Yes I think we did agree to move away from scaling (with a script ideally). But as I'm in there and want to reduce the mana cost, I added that change into this PR

src/WarcraftLegacies.Source/Setup/Spells/QuelthalasSpellSetup.cs
vampirika: Updated

YakaryBovine:
> > image We should avoid normalizing mana costs bit-by-bit like this, and instead either do it all at once or not at all. Otherwise we'll end up with a situation where some abilities have scaling mana costs and others do not, and the distinction between them won't be for any reason. And we'll put way more effort into it than we need to (we can just write a migration script!),
> 
> > 
> 
> > Other than that, looks good to me. There's nothing I really feel the need to test since I know you're on the ball about testing tooltips.
> 
> 
> 
> Yes I think we did agree to move away from scaling (with a script ideally). But as I'm in there and want to reduce the mana cost, I added that change into this PR

Reducing the mana cost is all good but the scaling should stick around until we can action changing over to a non-scaling world wholesale.

YakaryBovine:
Moving this back to In Progress as per your request on Discord!

mapdata/WarcraftLegacies/AbilityData/Skin/925905217.json
```diff
@@ -14,7 +14,7 @@
       "Level": 1,
       "Id": 828536161,
       "Type": 3,
-      "Value": "Gives 50% additional armor to nearby friendly units."
+      "Value": "Gives 50% additional base armor to nearby friendly units."
```
YakaryBovine, line 17: I feel like this probably doesn't say what you intended it to - this says it increases _base_ armor, which it doesn't. I'd just leave it as is; vanilla WC3 tooltips (ike Endurance Aura) don't explicitly say they increase by a percentage of the base and I don't it adds anything for us to either.

YakaryBovine, line 17:
```suggestion
      "Value": "Gives 50% additional armor to nearby friendly units."
```

YakaryBovine (changes requested):
One teeny tiny reversion to do, otherwise looks good.

mapdata/WarcraftLegacies/AbilityData/Skin/925905217.json
vampirika: This was a request from Techno, as a unit with 0 base armor but upgrades resulting in 2 armour will not gain any benefit from this aura.

YakaryBovine:
```suggestion
      "Value": "Increases the armor of nearby friendly units by 50% of their base armor, not including any upgrades."
```

YakaryBovine (approved):
Looks good

PR #3081, opened 2024-06-18
Technopig/share all command

Technopig1992:
"does what it says on the tin"


src/MacroTools/Commands/Share.cs
```diff
@@ -23,20 +24,43 @@ public sealed class Share : Command
     /// <inheritdoc />
     public override string Execute(player cheater, params string[] parameters)
     {
+      var cheaterTeam = cheater.GetTeam(); // Declare cheaterTeam once at the beginning
```
YakaryBovine, line 27:
```suggestion
      var cheaterTeam = cheater.GetTeam();
```
You can get rid of all these comments I reckon, it's clear what your code does just from reading it.

src/MacroTools/Commands/Share.cs
```diff
@@ -23,20 +24,43 @@ public sealed class Share : Command
...
+          {
+            SetPlayerAlliance(cheater, faction.Player, ALLIANCE_SHARED_CONTROL, true);
+          }
+        }
+
+        return $"Shared control with all factions on your team.";
+      }
+
+      // If not 'all', then proceed with previous share logic from Yak's code
+      if (parameters.Length == 0 || parameters[0].ToLower() == "")
```
YakaryBovine, line 46: You can get rid of this whole if statement. The intention is that `CommandManager` handles making sure the user provides the minimum set of parameters.

YakaryBovine (changes requested):
Looks good, just a few cleanups needed.
Also this is a feature so it needs to be re-pointed at `dev`.

Technopig1992:
> Looks good, just a few cleanups needed. Also this is a feature so it needs to be re-pointed at `dev`.

awesome , and didn't realise about CommandManager, ill remove it as suggested ty! and oops I made two by mistake and seemed to off kept the wrong one open! ill redirect

PR #3082, opened 2024-06-18
Slipstream portal for alterac fixed

SebastianTanoni:


YakaryBovine (approved):
We can merge this in because it does seem to make the Alterac portal a little more reliable, but the bug is still there so we'll keep that open. I'll put replication steps on the original Issue - they're definitely something a player would do!

PR #3083, opened 2024-06-18
Update Share.cs

Technopig1992:
removed the IF statement and unneeded annotations , merged into Dev


YakaryBovine (approved):
Looking good. Don't forget to add it to the 4.0.0 release notes.

PR #3105, opened 2024-06-28
Slipstream done, finally

SebastianTanoni:


YakaryBovine:
Awesome investigation work. I went with a different solution in #3110 but I used the findings here to figure out what needed to be done.

PR #3113, opened 2024-06-29
Updated Slumbering King quest and functions

vampirika:
- Extracted IsUnitBoat function to UnitExtensions.cs
- Updated ObjectiveAnyEnemyUnitInRects.cs to use Func
- Updated Slumbering King quest to use new Func


YakaryBovine (changes requested):
Looks great and seems to work.
Only change needed is that the Description of the Objective needs to reflect that it's non-boat units only. I would pass in a string parameter to the objective that describes the eligibility condition, i.e. in this case "boat". So Description would become `Description = $"Enemy {unitDescriptor} unit has entered {rectName}";`

vampirika:
> Looks great and seems to work. Only change needed is that the Description of the Objective needs to reflect that it's non-boat units only. I would pass in a string parameter to the objective that describes the eligibility condition, i.e. in this case "boat". So Description would become `Description = $"Enemy {unitDescriptor} unit has entered {rectName}";`

Think I've updated that

YakaryBovine (approved):
Looking good!

PR #3116, opened 2024-06-30
Added flight requirement to sapphiron quest

vampirika:


src/WarcraftLegacies.Source/Quests/Scourge/QuestSapphiron.cs
```diff
@@ -33,7 +34,7 @@ public sealed class QuestSapphiron : QuestData
       "Sapphiron has been slain, and has been reanimated as a mighty Frost Wyrm under the command of the Scourge.";
 
     /// <inheritdoc/>
-    protected override string RewardDescription => "Unlock one of the requirements to train Frost Wyrms. " +
+    protected override string RewardDescription => "Unlock ability to train Frost Wyrms. " +
```
YakaryBovine, line 37: While you're in there can you change this to the standard format, i.e. "Learn to train x from the y", and use GetObjectName for x and y?

vampirika: Updated

YakaryBovine:
Just on my phone so might be viewing things wrong, but I think you need to revert those war3mapmisc.txt changes. Not sure why those happen, it's happened to me a couple of times.

vampirika:
> Just on my phone so might be viewing things wrong, but I think you need to revert those war3mapmisc.txt changes. Not sure why those happen, it's happened to me a couple of times.

Done, dunno why that happened/happens

YakaryBovine (approved):
Looks great!

PR #3120, opened 2024-07-01
FixingDreadlordSleepnotworkingonhero

SebastianTanoni:


YakaryBovine (changes requested):
This clearly fixes the one instance of Sleep that was reported, but as discussed on Discord just now we need to check all instances of Sleep. Any of them that were using the default Targets Allowed would have been affected by the Blizzard change.

PR #3133, opened 2024-07-31
Adding new ambiants

SebastianTanoni:
To be merged after elementals


YakaryBovine:
Just marking these as draft until #3134 is merged, it's unreviewable before then due to file count anyway.

PR #3134, opened 2024-07-31
Cthun cleanup

SebastianTanoni:
to be merged after elemental objects


YakaryBovine:
Just marking these as draft until #3134 is merged, it's unreviewable before then due to file count anyway.

Issue #2761, opened 2024-03-18
Revisit fel horde demon gates


vampirika:
Forced options for demon gates removes player agency, so the units spawned from gates should be buffed / changed to offer valid options.
Potentially lower cost of gates early, higher late OR keep higher cost of demon gate to promote later use of gates and reduce unit lag.

YakaryBovine:
@vampirika Important to note that the change to limited options was made to fix a bug: https://github.com/AzerothWarsLR/WarcraftLegacies/issues/2034

The bug was that the limit on gates could be ignored by upgrading multiple gates simultaneously. If we revert that in any sense, the bug will need to be tackled a different way.

PR #3145, opened 2024-08-29
Vampirika/demon gates balance

vampirika:
Addresses #2761.
Simplified gates. Now only 2 gates with 2 options (Fel and Void).
Fel - Hound Pack, Fel Guard, Pit Fiend.
Void - Nether Drake, Voidlord, Terrorguard.

Gate type can take a parameter of maximum unit limit (default 12) per gate. For example, terrorguard now has limit of 6 per gate resulting in a maximum limit of 12 if both gates are upgraded to terrorguard. [Addresses limit bug]
Improved formula to account for multiple unit spawn e.g. hound pack with limit of 12 will not spawn 3 if there are 10 or 11 hounds alive for the associated gate.
Made gates start with half mana on upgrade as progress was wiped on upgrade.
Some balance changes.


src/MacroTools/Mechanics/DemonGates/DemonGateBuff.cs
```diff
@@ -85,10 +85,11 @@ public DemonGateBuff(unit caster, int demonUnitTypeId, float spawnInterval, int
     public override void OnApply()
     {
       Target
-        .IssueOrder("setrally", Target.GetPosition())
+        .IssueOrder(852002, Target.GetPosition())
         .SetMaximumMana((int)_spawnInterval)
         .AddAbility(_toggleAbilityTypeId)
         .IssueOrder("immolation");
+      Progress = _spawnInterval / 2;
```
YakaryBovine, line 92: I think instead of this workaround for mana getting wiped out, you could just fix it. With your fix to the `OnStack` method, you could just have it so that the new buff gets it Progress set to the value of the old buff, thus preserving Progress.

src/MacroTools/Mechanics/DemonGates/DemonGateBuff.cs
```diff
@@ -85,10 +85,11 @@ public DemonGateBuff(unit caster, int demonUnitTypeId, float spawnInterval, int
     public override void OnApply()
     {
       Target
-        .IssueOrder("setrally", Target.GetPosition())
+        .IssueOrder(852002, Target.GetPosition())
```
YakaryBovine, line 88: It's great that you've switched over to the ID version of this (ideally I'd like to delete the string version once all uses of it are gone), when we only have a raw ID to supply I think we want to maintain readability by using theconversion, so like (I don't remember the actual conversion function off the top of my head).

What would probably be even better is having an `OrderConstants` static class that contains these. It would, I think, only need to contain the weird ones like Set Rally, since `Constants.cs` already generates them for all abilities.

`.IssueOrder(ToOrderId("setrally"), Target.GetPosition())`

src/MacroTools/Mechanics/DemonGates/DemonGateBuff.cs
```diff
@@ -130,10 +131,11 @@ private void SpawnDemon()
     {
       for (var i = 0; i < _spawnCount; i++)
       {
-        var spawnedDemon = CreateUnit(Target.OwningPlayer(), _demonUnitTypeId, SpawnPoint.X, SpawnPoint.Y,
-          Target.GetFacing() + FacingOffset)
-          .IssueOrder("attack", RallyPoint);
+        var spawnedDemon = CreateUnit(Target.OwningPlayer(), _demonUnitTypeId, SpawnPoint.X, SpawnPoint.Y, Target.GetFacing() + FacingOffset)
+          .IssueOrder(851983, RallyPoint);
```
YakaryBovine, line 135: Refer to my other comment on using `IssueOrder`

src/MacroTools/Mechanics/DemonGates/DemonGateType.cs
```diff
@@ -20,11 +21,13 @@ public sealed class DemonGateType : PassiveAbility
     /// <param name="demonUnitTypeId">The unit to spawn.</param>
     /// <param name="spawnInterval">How often to spawn the unit.</param>
     /// <param name="spawnCount">How many of the unit to spawn each time.</param>
-    public DemonGateType(int gateUnitTypeId, int demonUnitTypeId, float spawnInterval, int spawnCount) : base(gateUnitTypeId)
+    /// <param name="spawnLimit">How many of the unit to spawn each time.</param>
+    public DemonGateType(int gateUnitTypeId, int demonUnitTypeId, float spawnInterval, int spawnCount, int spawnLimit = 12) : base(gateUnitTypeId)
```
YakaryBovine, line 25: Just take out the default IMO. The caller should know what limit they're looking for - there's not many to configure.
```suggestion
    public DemonGateType(int gateUnitTypeId, int demonUnitTypeId, float spawnInterval, int spawnCount, int spawnLimit) : base(gateUnitTypeId)
```

src/MacroTools/Mechanics/DemonGates/DemonGateType.cs
```diff
@@ -20,11 +21,13 @@ public sealed class DemonGateType : PassiveAbility
     /// <param name="demonUnitTypeId">The unit to spawn.</param>
     /// <param name="spawnInterval">How often to spawn the unit.</param>
     /// <param name="spawnCount">How many of the unit to spawn each time.</param>
-    public DemonGateType(int gateUnitTypeId, int demonUnitTypeId, float spawnInterval, int spawnCount) : base(gateUnitTypeId)
+    /// <param name="spawnLimit">How many of the unit to spawn each time.</param>
```
YakaryBovine, line 24: Don't think this comment is right!

YakaryBovine (changes requested):
Nice, looks like a huge improvement. I think the comments I left are helpful but I haven't launched this in game.

FWIW this definitely not something I'd merge into `main` as it will have all the usual issues of conflicts with `dev` and will draw priority away from the work we were focusing on in `dev`. But, I can see the desire to actually release something while Old Gods are on indefinite hold. Just something you might want to check in with Sebas about.

src/MacroTools/Mechanics/DemonGates/DemonGateBuff.cs
vampirika: I tried that but I couldn't seem to get it to work, presumably because the new Buff has new progress.

src/MacroTools/Mechanics/DemonGates/DemonGateType.cs
vampirika: Amended

src/MacroTools/Mechanics/DemonGates/DemonGateType.cs
vampirika: Amended

src/MacroTools/Mechanics/DemonGates/DemonGateBuff.cs
vampirika: Amended - added OrderId("setrally")

src/MacroTools/Mechanics/DemonGates/DemonGateBuff.cs
vampirika: Amended, as above

YakaryBovine (approved):
Looks good to me

src/MacroTools/Mechanics/DemonGates/DemonGateBuff.cs
vampirika: Yeah, tried having another go at it but would need some help to look at it which I don't think either of us want to put in. I think this is an acceptable solution in terms of gameplay. Although having it reset to 0 wouldn't be game breaking either, having something here _might_ remind us to look into it later.

PR #3175, opened 2024-10-11
Vampirika/bugfix and old god balance

vampirika:
Updated names for black empire heroes in code (which was affecting a quest text).
Lowered attack and health of Cthun and Nzoth.
Added attack/defense upgrades to black empire casters and lowered attack damage (in line with other factions).
Lowered cost of black empire elite form 90 to 75 (in line with other factions).
Removed mention that hideous appendages has same attack as old gods (as it doesn't) and adjusted tentacle damage.
Moved beastary in orgrimmar that produced units that were stuck.


mapdata/WarcraftLegacies/AbilityData/Skin/1261449537.json
```diff
@@ -63,7 +63,7 @@
       "Level": 1,
       "Id": 828536161,
       "Type": 3,
-      "Value": "This unit is surrounded by Tentacles which attack nearby enemy units. Each Tentacle has the same attack damage as the hero."
```
YakaryBovine, line 66: Oops... This was definitely something I had meant to program, my bad. Maybe one day I'll put it back in lol.

YakaryBovine (approved):
Looks great

PR #3185, opened 2024-10-21
Vampirika/bug fixes gilneas stormwind

vampirika:
### Release notes:
Fixed gilneas starting town hall unable to perform actions.
Stormwind marksman starting mana set to maximum.
Stormwind marksman arrows now poison themed to match effect with a DoT and upgrade from spearman throwing poison spears.
Green dragons can now properly attack all targets after emerald flame upgrade.


YakaryBovine (approved):
Looks good to me!

PR #3205, opened 2025-01-10
Scourge Hivemind rework

YakaryBovine:
**Domination** 
- Added a new Power, Domination, which allows the Scourge to train and control mindless Undead if and only if they have it
- The Frozen Throne is no longer destroyed upon reaching 0 hit points; instead, it becomes Ruptured, turning off Domination and removing all of its abilities permanently
- The Throne can be recaptured by completing The Ascension, giving Arthas the Crown and turning Domination back off, but the Throne still cannot cast abilities _and_ it can now be permanently destroyed
- If Arthas permanently dies as the Lich King, the Crown cannot be returned to the Throne
- The Throne can _also_ be vacated if Scourge leaves the game, allowing it to be destroyed so the Crown can be retrieved
- Some cool dialogue is played when Arthas is alive and the Throne is fractured
- Fixed Throne model

Todo:
- Adjust Frostwolf's quest to kill Frozen Throne to account for the inability to destroy it without Scourge completing The Ascension
- Add a way for Scourge to decide to go Lich King without them needing control of the Throne (because they don't have control when it's fractured)
- Switch out model when the Throne is vacated


src/WarcraftLegacies.Source/Factions/Scourge.cs
```diff
@@ -179,11 +202,30 @@ private void RegisterDialogue()
               EligibleFactions = new List<Faction> { this }
             }
           }));
-    }
 
-    private void RegisterHelmOfDominationLogic()
-    {
-      HelmOfDominationDropsWhenScourgeLeaves.Setup(this, _artifactSetup.HelmOfDomination, _allLegendSetup.Scourge.TheFrozenThrone);
+      TriggeredDialogueManager.Add(new TriggeredDialogue(
```
Technopig1992, line 206: This dialogue did not play during my tests.

Technopig1992:
- Throne model starts the game facing an odd direction. 
- Scourge dialogue did not play when Throne was ruptured , whilst Arthas was alive. 

image

PR #3220, opened 2025-01-24
Technopig/progeny gland bal

Technopig1992:
Reduced total DPS by lowering attack base + speed.
Reduced Acquisition range from 1500>1000.
Reduced duration from 35-15.
Increased Summon duration from 10>15.
Increased Soldier gold cost 16>18.
Swarm Beetle can no longer target CPs.

Misc
Added Flight research for Dalaran


mapdata/WarcraftLegacies/AbilityData/Skin/843863129.json
```diff
@@ -29,7 +29,7 @@
       "Level": 1,
       "Id": 828536161,
       "Type": 3,
-      "Value": "The silithid spawns 3 Swarm Beetles upon death. They last for 35 seconds."
+      "Value": "The silithid spawns 3 Swarm Beetles upon death. They last for 15 seconds."
```
YakaryBovine, line 32: You don't have to fix it but... data fields!!!!!! This is not a tooltip we should have to update manually

PR #3234, opened 2025-02-03
Implemented ability for multiple factions within a faction choice to share the same starting location

Technopig1992:
- It is now possible to specify the unit category of any given registered object
- Factions with a Faction Choice can now share the same preplaced units, which can be swapped out for equivalent unit types from another faction if a different one is picked


YakaryBovine:
@Technopig1992 If this is ready, can you publish it and fix that build error?

PR #3235, opened 2025-02-03
Technopig/FactionCleanUp Part 2

Technopig1992:
# Faction Clean Up 
## Changes 

- A single Goldmine can now be registered by multiple factions and will not be removed once a Faction is picked. 
- Goldmines are now Removed rather than Killed when a Faction is Removed to avoid any unwanted SFX.


YakaryBovine:
Closing in favour of simpler solution here: https://github.com/AzerothWarsLR/WarcraftLegacies/commit/8a00354793c5135b0884d78fff7bd78ba1313483

PR #3261, opened 2025-03-02
New design of The Nexus - Zbovo

Technopig1992:
Add new abilities and modify data for elementalist and judge.

- Duplicate PR , reviewed and approved


SebastianTanoni:
image
 berserk B not capitalized. 
 remove 1 "and" from the phrase. "This unit is Berserk; it will deal more damage,  absorb spell damage and block other damage."

SebastianTanoni:
image
Sunfury already has a unit called Arcane Ward that does something different. The new nexus one will need to be called something else

SebastianTanoni:
image

This model should not be used for the arcane ward, it will get confused with a spell effect. The arcane ward should have a model that clearly indicates this is a ward.

SebastianTanoni:
image

Need to remove this from Abolish magic name (Arcane Ward)

SebastianTanoni:
image

The description should read:

"It has the Feedback and Abolish Magic abilities."

SebastianTanoni (changes requested):
Good work overall, some changes to do.

SebastianTanoni:
image

image


Im almost certain these icons are used in existing spells already.

zbovo:
> image
> 
> image
> 
> Im almost certain these icons are used in existing spells already.

hmmI checked it before using it. It seems that they only exist but are not applied to abilities. Can you tell me what abilities they are used for?

SebastianTanoni:
> > image
> > image
> > Im almost certain these icons are used in existing spells already.
> 
> hmmI checked it before using it. It seems that they only exist but are not applied to abilities. Can you tell me what abilities they are used for?

image

A quick search and i found this spell.

Always assume that the icons are used somewhere. You should delete unused icons. When making new mechanically unique spells always assume you will need a new icon

PR #3292, opened 2025-04-11
Technopig/Exodar Slipstream portal fix

### LINKED ISSUE #3283: Dimensional Generator has no inventory slots

Technopig1992:
Refactored Slipstream portals to use a random point in a region rather than an exact point .
This fixes a bug where Exodar's Slipstream portals would not work after being cast more than once.

closes #3283


vampirika (approved):
Assuming you've tested the functionality in game, looks good to me.

Might this be something to consider using for other slipstream portal issues (.e.g KT and legion)?

Technopig1992:
Tested and working correctly. 

I think the issue with the other portals was unrelated to be fair, something to do with the destination portals crossing over.

PR #3304, opened 2025-04-20
Nerf Nexus

zbovo:
1.Decrease the cast range of Ice Explosion 650 to 500 The interval changes to 0.5/0.45/0.4s
The damage of every blast 100→80
2.The AOE of Dragon Guard 600→500
Mana regen changes to 0.4/0.6/0.8/1
Can only affect your own units now
3.The mana maximum of judicator 600→500
4.Delete the devour magic of crystal lord and reset its basic attributes,gold cost change to 75. 5.Reduce Jaina intellectual growth by 1.
6.Add a Hotkey V for Self Destruct


Technopig1992 (approved):
Just number changes and tooltip changes seem correct so approving.

PR #3305, opened 2025-04-20
Hotkey Migration added

Technopig1992:
All Unit training hotkeys and ability casting hotkeys are now auto set based on there button position. Hero Learn hotkeys and Research hotkeys are excluded and still require manual adjustment.

### DIFF
--- a/src/Launcher/MapMigrations/HotkeyMigration.cs
+++ b/src/Launcher/MapMigrations/HotkeyMigration.cs
@@ -1,0 +1,98 @@
+﻿using System;
+using War3Api.Object;
+using War3Net.Build;
+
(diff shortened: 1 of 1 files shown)


Technopig1992:
Have tested and all seems fine but worth another couple pairs of eyes I reckon just to be safe,

vampirika (approved):
Looks good, but from what I can tell custom keys still overides it.

So for example, my issue with thrall in frostwolf is shackles will still have the Q hotkey (from aerial shackles on dragonhawk) which conflicts with the Q of forked lightning.

Still think it is worth adding though, means we won't have any issues in hotkeys if people use classic

PR #3307, opened 2025-04-21
No Target Death Pact

Technopig1992:
## Death Pact Changes

In an effort to make Scourge, a two man faction , less APM intensive. I made the following changes to Death Pact to give the player one less thing to do.

- Replaced Death Pact and Dark Ritual with coded versions.
- Both abilities will now select there target automatically, killing the unit and converting the hitpoints into health and mana.
- Targets are prioritised in the following order:
                                                    - Highest level summoned unit.
                                                    - Summoned unit.
                                                    - Living unit.


vampirika (approved):
Seems fine to me

Technopig1992:
> Seems fine to me

Good good! , ill merge it but one to keep an eye on for balance reasons as it might just lead to unkillable DKs xD can lower % values if needed later.

PR #3308, opened 2025-04-22
Legion Heroes rework

vampirika:
### Reworked Legion heroes (Dread lords).
Removed invisibility.
Added a new passive ability "Vampiric Siphon" which restores a flat amount of health to the hero every time it damages an enemy unit. This is to incenvise using the heroes in melee combat.
Added quest rewards to 3 different quests that boost the strength of each hero's passive.
Added multiple new spells and moved spells around between dread lords.

Malganis (Intelligence)
Q - Carrion Swam
W - Parasite
E - Brilliance aura
R - Inferno

Tichondrius (Strength)
Q - Touch of death -  changed finger of pain into a melee range forked lightning.
W - War stomp
E - Vampiric aura
R - Inspire madness

Anetheron (Agility)
Q - Shadow Strike
W - Windwalk
E - Pulverize
R - Doom


Technopig1992:
Looking good man I like it! , just some random clean up stuff , let me know if there's something up with the actual file for the DISBTN for Siphon and I can do it again. 


Touch of Death tooltip typo.
image

Pulverise Icon for Hero Learn should by BTN not PASBTN. 
image

Tooltip Name needs Level 1,2,3,4 
image

Parasite and Carrion Swarm share Learn Hotkeys, this is sometimes due to having the same Learn Icon Location if not the actual hotkey.
image

Vampiric Siphons DISBTN doesn't seem to work properly 
image

No effect seems to trigger when Vampiric Siphon triggers. 
It doesn't seem like Pulverise damages friendly units like vanilla Pulverise so the tooltip should mention enemies.  
Vampiric Siphon's tooltip states it scales with hero level but I think it only scales with ability level.

Technopig1992 (changes requested):
clean up points mentioned in comments, consider changing effect to heal over spirit touch for consistency

vampirika:
> clean up points mentioned in comments, consider changing effect to heal over spirit touch for consistency

Pulverize damages allies? :o I'll probably just update tooltip I think, don't think it should damage allies.
Didn't seem to be a BTN for pulverize though?

I didn't put a SFX for vampiric siphon, do you think it's needed? Would be every auto and spell tick.
I imported the DISBTN but dunno if I need to put it somewhere specific?
It does scale with level but only +1 per level for now - do you think more? +2 per level maybe?

Others np, will fix them

Technopig1992:
> > clean up points mentioned in comments, consider changing effect to heal over spirit touch for consistency
> 
> Pulverize damages allies? :o I'll probably just update tooltip I think, don't think it should damage allies. Didn't seem to be a BTN for pulverize though?
> 
> I didn't put a SFX for vampiric siphon, do you think it's needed? Would be every auto and spell tick. I imported the DISBTN but dunno if I need to put it somewhere specific? It does scale with level but only +1 per level for now - do you think more? +2 per level maybe?
> 
> Others np, will fix them


Yeah on Bonebreakers for example , its just the default I think but deffo odd. 
I thought I saw spirittouch for it but might of been looking at something else, but also Heal I reckon is fine , like its always going off anyway, I also think it does for Vamp Aura to maybe? can leave it off if you want either way.
With the DISBTN it doesnt need to go anywhere but needs the correct path , so the same name as the normal BTN but with the DIsabled path , check another DISBTN in asset manager to get it. 
Oh no that's cool I just didnt realise , like you said before be hard to balance , so at level 20 they get 25 healing per unit damaged? We can just keep an eye on it. 
There prob isnt a normal BTN for Pulverise atm as I dont think its used as a hero ability , the hero version is pretty much just cleave, but that does a % of attack dmg rather than a set amount of only hits enemy units - I can also make one if you need.

vampirika:
Added the correct BTN and DISBTN for vampiric siphon and pulverise which both look good now.
Ah I read SFX as sound effects - Changed the special effect to use heal.

Unless we gonna simulate huge fights to test balance I'd say ready to go?

PR #3314, opened 2025-05-10
Elemental Lord

zbovo:
rework all abilities,and fix some bugs


Technopig1992:
image

Need to add to the reward description for Druids Demise that it unlocks the ability for Elemental Lord.

image
Same with Subduing Neptulon

image
Needs to be a Passive Icon rather than an Active Icon 

Tooltips 

Core of the Gale: 
Need to change the name and use Ward instead of Core , as it is summoning a ward/totem. Below is just a suggestion feel free to amend the name etc.  

Wind Ward / Gale Ward / Ward of Wind / Ward of Gale / Stormcaller Totem / Gale Totem / Windcaller Totem  
"Summons an enchanted ward that empowers nearby allies, increasing the movement speed of up to two friendly units by 10% and their attack speed by 20% every 3 seconds. While the ward exists, all Elemental Lords gain 50% evasion and slows enemy units within 400 range by 15%. Lasts 45 seconds."


Earth Protection:  
I get the idea , name cold be better though , Rock Shield / Seismic Guard / Stone Shield , again just suggestions.

"Increases the armor of all nearby Elemental Lords by 20 and increases this units health regeneration by 50.|n|cffffcc00Lasts 15 seconds." - assuming the health regen only works on the casting unit not all? 

Water Prison: "Each attack has a 40% chance to trap the target in a binding prison for 20(3) seconds, dealing 30 damage per second. A unit can only be affected by Water Prison once every 8 seconds.



Balance:

Water Prison proc rate is far too high , you have 6 of these units this procs each time they collectively attack anything , it working on heroes too is very unfair, I would make it not effect heroes and drastically reduce the proc rate . bear in mind you have also given them, Incinerate on each attack on top. 

Earth Protection: Cool spell but the health regen amount is far too high , Burrow and Gargoyle Stone Form provide 8 hp regen , this provides 50, needs to be much lower 
 
Core of the Gale evasion at 50% seems very high for a ranged unit , maybe consider lowering it .



All very nice work! cool theme , all fits together nicely and really nice looking icons and visuals . Really well done here mate! I will knock up a ranged storm surge for you to add in .

zbovo:
Basically complete, there should be no problem now
1. Core of the Gale is renamed as Ward of Gale, and the skill data is corrected.
2. The icon of Water Prision is replaced, the probability is reduced from 40% to 30%, and the cooldown is increased from 8 seconds to 10 seconds.
3. The health restoration of Earth Protection is reduced from 50 to 30.
4. The probability of Mass Ensnare is increased to 75%, and the range is increased to 250.
5. Add an instruction for the task of Skywall to unlock the skills of Elemental Lord.
6. Strengthen the water element summoned by Tidal Lord, add a new skill, and reduce the summoning quantity from 3 to 2.
7. The damage of Incinerate of Elemental Lord is reduced from 40 to 30, the attack type is corrected to Hero Attack, and the attack interval is increased to 1.7.
8. The mana cost of Storm Surege is modified to 90.
9. Fix the hotkey and key position issues of Farstrider and Tidal Lord research.

Technopig1992:
Top work man! 

All looks good, I changed Storm Surge into a ranged ability but lacked the permissions on Github to push directly to your PR so have had to create a new one based off of it.

Technopig1992 (approved):

PR #3320, opened 2025-05-14
Rework Akama

zbovo:
Q ShadowAssault
W shadow aura
E  Crippling Strike
R  Vanish


Technopig1992:
Really nice work , you really nailed the brief spot on! All the abilities are thematic for Akama and a "shadowy assassin".

image
No problem , and jus try to avoid using orb effects as for w/e in WC3 they just have there own hard coded rules. You can change it to a custom Critical Strike and get a similarly desired effect is using 100% rate. 

## Balance 
image
This ability is super cool however with all the elements combined its very overpowered.  The fact it can reset the CD means that it is very easy to use just on repeat to any backline unit in the fight. I would say keep all the elements but please lower the thresholds. Reduce the Execute threshold to be a rarer occurrence and keep it consistent across the levels please.  The attack speed increase is also far too high, just lower these considering the ability still has 3 more elements. The single target damage is also spell damage so ignores armor , so also please lower these values by about 25% each level. 

image
Again please lower the values used here, I am aware the radius is much lower but these have higher values now as a normal ability than it did as a level 3 ultimate, levels 1 and 2 are fine but please reduce 3 and 4. 

image
Nice cool ability , see note above regarding abilities based on Orbs and change accordingly. (Balance this one with the attack speed increase of Shadow Assault , having a unit attack 250% quicker whilst he attacks do nearly double damage just results in insane dps. 


## Tooltips
Shadow Ambush
Akama strikes from the shadows, entering a shadow state and gaining increased attack speed. He charges toward a target unit, phasing through other units and dealing damage upon impact. If the target’s HP is below 20% and it is not a hero, Akama executes it, instantly refreshing Shadow Ambush.

Shadow Aura 
This Hero’s presence weakens nearby enemies reducing their attack damage

Crippling Strike 
Akama’s attacks tear through his enemies, rending their flesh and making them more vulnerable. All units and heroes he attacks suffer increased damage.

thats legit it , just balance nothing else to change! again really nice job man!

Technopig1992 (changes requested):
See conversation

zbovo:
1.Nerf shadow aura,now just 50% at lvl 4,but add evasion for every lvl （20% 25% 30% 35%）
2.The execute threshold is fixed at 25%
3.Code crippling strike,keeping the original value
4.every tooltip fixed
5.Buff the blink max distance by 100

Some personal ideas
I don't think this hero can be compared to any other. Sylvanas and Akama have the same primary attribute, yet their survivability varies significantly. Sylvanas is relatively easy to operate and boasts high DPS upon reaching level 10. However, after over 20 tests on Akama, I've found that once he charges in, he can only deal damage for a maximum of 2 seconds before getting killed. If we must make comparisons with other heroes, their power levels need to be factored in.
 
Take Illidan for example. His Q ability can boost attack speed by 400%, which can't be dispelled. He also has a transformation skill and a W ability for significant health regeneration. But in practice, his performance is just so - so. He can only shine when equipped with numerous powerful items. Also, his passive deals area - of - effect damage.
 
In contrast, Akama's Q ability is extremely difficult to use. You have to constantly target low - health units. It's tough to manage even with just a few units, let alone in large - scale battles. Moreover, he has no means of self - healing. Are we supposed to rely on Blink? I think Wind Walk is far superior to Blink.
 
If Akama had Wind Walk, I'd admit that when combined with Q, it might be overpowered and allow him to assassinate many heroes. But currently, he doesn't have it. He's an assassin out in the open. There's no way he's overpowered. He can only be strong at level 10, and the conditions for reaching that are extremely demanding.
 
So, my stance is that we shouldn't just focus on numerical values; we need to consider actual performance. That's my personal opinion.

At last,If you don't agree with my point of view, you can nerf it at any time. Remember to update the tooltip. This is just my personal opinion.

Technopig1992:
Im not making comparisons with heroes im stating - im comparing abilities and when Shadow Assault does more DPS than other single target damage abilities,  the same level as attack speed boost as other albitites, execute on top and the chance to reset CD , when I say its overpowered im not plucking it from thin air

PR #3322, opened 2025-05-15
Madsen/chen replacement

Ripzss:
Rokhan replacing Chen Stormstout


Technopig1992 (changes requested):
## Bug(s)
- Need to open in world editor and save the map again and push to update the minimap (currently is a black square) 

## Balance
Serpent Wards - are the other Wards like this spell immune too? seems a bit broken as they scale with attack and defence upgrades and you get a large amount of them , wards being the lowest prio for units to attack on top. If Voljins are like this are haven't had any complaints maybe its fine. 

But thats it , cool hero !

Technopig1992 (approved):

PR #3324, opened 2025-05-18
Technopig/phantom step 

Technopig1992:
Replaced Wind Walk on Dreadlords + Anetheron with Phantom Step.

Phantom Step

Creates a mirror image of the caster to deceive enemies while the casting unit becomes invisible, moving <VP09,DataB1,%>% faster for <VP09,Dur1> seconds.


zbovo:
image
image
these 2 wrong tooltip,but work correctly.

zbovo (approved):

PR #3325, opened 2025-05-18
Remove

zbovo:
- Remove Magic Sentry from all Towers. ( this is pretty annoying as its on all versions of the tower D:)
- Remove Reveal from all unlimited buidlings. (Towers etc)
- Create a new Tower for Quel'thalas called Sunwell Tower (or something) that has increased scale and HP and preplace 4 at the Sunwell (theres spots to place them where the apth ends in each corner of the well). These Towers should have both Reveal and Magic Sentry.
- Add Magic Sentry to the following capitals; Capital Palace, Stratholme , Violet Citadel, Tyrs Hand, Silvermoon and Windrunner Spire.
- Add Magic Sentry and Reveal to Kirin Tor Towers.


Technopig1992:
Nice work , ty for doing this. Have replicated in a new PR with a change.

PR #3341, opened 2025-05-25
Technopig/DynamicFactionColours 

Technopig1992:
### Changes 

Each faction is given three prioritised colours. If a colour is already in use, the next available one is selected. If all three are taken, a different unclaimed colour is assigned. When a faction is defeated or leaves, its colour is released and made available for reassignment.

- FactionColourManager class created.
- UI Hex Colours are now linked with player colours. 
- Each faction class updated and have three colours assigned.
- Updated logic in Faction class for Player Colour assignment and Defeat logic.

### DIFF
--- a/src/MacroTools/FactionSystem/Faction.cs
+++ b/src/MacroTools/FactionSystem/Faction.cs
@@ -86,23 +86,10 @@ namespace MacroTools.FactionSystem
       Id = _highestId + 1;
       _highestId = Id;
 
-      
-      foreach (var color in colorPriority)
-      {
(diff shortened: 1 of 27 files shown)


Technopig1992:
image
\\ To Do
Need to add a method that applies the same faction colour to the intro quest text . This is currently done via hex strings .

Technopig1992:
@vampirika could you review, approve and merge this into Dev when you have a chance then we can start merging main and dev

vampirika (approved):

PR #3348, opened 2025-06-03
Energy wave rework

Technopig1992:
### Changes 

- Fully coded , base ability accounts for 0 healing and jumps , acts as an ability placeholder. 
- Bounces up to 3 times per wave , diminishing returns applied , 10% reduction per bounce.
- A buff is applied to the last healed unit in the wave, if a unit dies with this buff it will cast a new wave on death to nearby units, this will repeat until the buff is missed. These waves have a 10% diminish rate. 

## Teron

- Same ability with higher base heal numbers. 

## To Test

- Ensure the buff is visible on the unit. 
- Ensure buff is removed.
- Check heal amounts.
- Icons, tooltips etc


Technopig1992:
hotkey Teron 
second wave units effected cap at 3.

PR #3366, opened 2025-06-28
Technopig/DjinnEfreetSpells

Technopigg:
### Removed Abilities

- Endurance Aura removed from Efreet.
- Thorns Aura removed from Djinn.

### New Abilities 

- Whim of the Winds added to Djinn.
- Warped Malediction added to Efreet.

### Warped Malediction

Autocast ability that, when cast, will inflict a random debuff onto the target enemy unit. The debuff is randomly selected from a list of whitelisted buffs. 

**Current Debuffs:**
- Curse
- Soul Burn
- Mind Flay

### Whim of the Winds 

Autocast ability that, when cast, will bestow a random buff onto the target friendly unit. The buff is randomly selected from a list of whitelisted buffs.  

**Current Buffs:**

- Inner Fire 
- Roar , single target 
- Bloodlust


src/WarcraftLegacies.Source/Spells/WarpedMalediction/WarpedMalediction.cs
```diff
@@ -0,0 +1,68 @@
...
+        
+        foreach (var buff in _possibleBuffs)
+        {
+          if (GetUnitAbilityLevel(target, buff.AbilityId) > 0)
+          {
+            return;
+          }
+        }
+
+        var randomValue = GetRandomReal(0, 1);
```
vampirika, line 53: I think this will only get random values of 0 or 1, which means this logic won't work. I think you want

var randomValue = GetRandomReal(0, _possibleBuffs.Length - 1);
var selectedBuff = _possibleBuffs[randomIndex];

src/WarcraftLegacies.Source/Spells/WhimOfTheWinds/WhimOfTheWinds.cs
```diff
@@ -0,0 +1,56 @@
...
+      if (!CastFilters.IsTargetAllyAndAlive(caster, target))
+        return;
+
+      foreach (var buff in _possibleBuffs)
+      {
+        if (GetUnitAbilityLevel(target, buff.AbilityId) > 0)
+          return;
+      }
+
+      var randomValue = GetRandomReal(0, 1);
```
vampirika, line 46: Same as previous comment

src/WarcraftLegacies.Source/Spells/WarpedMalediction/WarpedMalediction.cs
Technopigg: it sets each at 0.33 something , it gives a slight favour to the first listed spell , but very slight.

When testing I di 20 casts and got a good mix

vampirika: Hmm okay. When I searched that random method it seemed to say it only got values 0 and 1, but if it works then all good

vampirika (approved):

PR #3371, opened 2025-07-13
Technopig/quest plague async timings

Technopigg:
An attempt to reduce the lag spike caused when the quest Plague completes


Technopigg:
Closing as no FPS change was observed

PR #3379, opened 2025-09-10
feat: add `csharp-to-lua` for transpiling C# to WC3 maps in folder mode

### LINKED ISSUE #3378: Proposal: Add Visual Studio Launch Profile to generate the map's script to the output folder
Since Warcraft 3 supports running maps directly from folders, I'd like to add a launch profile that compiles the map script and writes it to the "maps" folder. This would allow for faster iteration during script development without requiring full map packaging

alrun3:
Adds a `csharp-to-lua` command to transpile C# source code into Lua and write it into an existing Warcraft 3 map folder.

Workflow:
* Run "WarcraftLegacies: Convert MapData to W3X" to generate the base map folder
* Run "WarcraftLegacies: Transpile C# to Lua" whenever you want to update the map's Lua scripts from C#

Closes #3378


YakaryBovine (approved):
Great idea, this will be very helpful for iterating on changes that only tackle code.
Approved, but my only question is, how do you recommend the user runs the map in WC3 after doing this? Manually opening WC3 and navigating to the map folder?

PR #3388, opened 2025-09-13
🚧 Migrate from `War3Api` to `WCSharp.Api` and upgrade to latest version

alrun3:
> ⚠️ **Note:** Expect force-pushes and frequent rebases during the draft phase, as the implementation evolves.

This PR represents the endeavour to upgrade the core WC3 libraries by replacing `War3Api` with `WCSharp.Api` and updating to the latest available version.

It builds on the work done in #3375 and #3382, which is summarized in the implementation plan: #3385.

In its draft state, this PR serves as a compilable reference, running on the latest bits. It will be continuously rebased on `dev` as more parts of the implementation plan are completed.

The PR is intended to be a living point of reference, where we can iterate, review, and split off focused work into separate tasks or PRs until it's ready to merge.

The commits in this branch were cherry-picked from https://github.com/alrun3/WarcraftLegacies/tree/upgrade-wcsharp.

Closes #3373

PR #3401, opened 2025-09-16
🚧 Integrate `WCSharp.Dummies`to simplify dummy casting API

alrun3:
This PR replaces the internal dummy handling with WCSharp's DummySystem, removing custom recycling logic and centralizes casting logic in `DummyCaster.Cast(...)`, supporting:
* No target
* Point target
* Unit target (with optional facing and origin)

The new class uses `DummySystem.GetDummy(...)` and `DummySystem.RecycleDummy(...)` internally for dummy pooling and cleanup.

PR is based on: https://github.com/AzerothWarsLR/WarcraftLegacies/pull/3388


alrun3:
Just an update on this. We still can't proceed with this PR because there is a bug upstream that is not being fixed that makes this code 32 times worse than the current implementation: https://github.com/Orden4/WCSharp/pull/21

PR #3418, opened 2025-09-18
🚧 Migrate from `WCSharp.Api.Common` to native C# wrappers

alrun3:
This PR removes the dependency on `WCSharp.Api.Common` (Common.lua) and replaces it with native C# wrapper equivalents provided by `WCSharp.Api`.

Migration Approach

1. Remove `global using static WCSharp.Api.Common;` from global usings.
2. Apply a naive Regex-based transformer to replace Common.lua function usage with C# equivalents.
3. Iteratively fixed compile errors until the project built successfully.


YakaryBovine (approved):
Looks great. I've done a general skim over the file changes but not all 532 of course. Launched the map as Illidari and did a handful of quests, all seemed to work as before. The diff in the war3map.lua looks to only have sensible native changes.

PR #3433, opened 2025-09-22
chore: upgrade target framework to net9.0

alrun3:
Depends on https://github.com/AzerothWarsLR/WarcraftLegacies/pull/3432


YakaryBovine:
Out of curiosity, can we go straight to .NET 9?
Looks good to me anyway.

alrun3:
> Out of curiosity, can we go straight to .NET 9? Looks good to me anyway.

I would prefer that 👍

YakaryBovine (approved):
Awesome, glad to be up to date.

PR #3434, opened 2025-09-22
refactor: move `ORDER_` constants to shared layer

alrun3:
OrderIds aren't specific to WarcraftLegacies, so this PR proposes moving them to MacroTools.Shared for broader use


src/MacroTools/Extensions/UnitExtensions.cs
```diff
@@ -73,19 +73,6 @@ public static string GetProperName(this unit whichUnit) =>
   public static void Ping(this unit whichUnit, float duration) =>
     PingMinimap(whichUnit.X, whichUnit.Y, duration);
 
-  /// <summary>
-  /// Orders a unit to perform a specified order at a specified <see cref="Point"/>.
-  /// </summary>
-  [Obsolete("Use the version that takes an integer order ID instead.")]
-  public static void IssueOrder(this unit unit, string order, Point target) =>
```
YakaryBovine, line 80: Thanks for getting rid of these :D

PR #3436, opened 2025-09-24
refactor: use compound assignments and inline `RestoreMana`

alrun3:
Available with the WCSharp upgrade


YakaryBovine (approved):
Good stuff

PR #3437, opened 2025-09-24
refactor: prefer the `IDictionary.TryGetValue(TKey, out TValue)` method

alrun3:
Available with the CSharp.lua upgrade


YakaryBovine:
I'm excited for this!

Issue #3440, opened 2025-09-24
Decide on a C# language version strategy


alrun3:
We need to decide on a strategy for the C# language version we're targeting.

Currently, we're using .NET 9, which defaults to C#13, but the transpiler, CSharp.lua, only supports up to C#10. Despite this, there’s already some usage of C#11 features like `required` properties. These are accepted by the compiler but not supported by the transpiler. Targeting an unsupported language version leads to problems such as IDE suggestions for invalid features and the risk of writing code that compiles but fails during transpilation, or worse, at runtime.

Some options:

* Lock the language version to C#10 and remove the unsupported `required` modifier
* Use a higher language version (e.g. C#11) but maintain a rulelist of features that are not supported by CSharp.lua, and configure tooling to suppress suggestions for those features

At the moment, nothing enforces either approach, so it is easy to introduce code that looks valid but isn't

alrun3:
It would be simplest to just target C#10, but I'd also like to see the platform move forward where possible, especially when there are useful features in later versions that don't break the mapscript.

My current motivation is to bring the analyzer events down to zero. I think there were around a thousand when I checked last night.

Shall we go with option 2 and try to curate the list? Targeting C#11 initially should keep things manageable, and we can always iterate on new findings if we don't catch everything up front, or reconsider if it turns out to be too much overhead.

I'm happy to take this on and put together a draft PR for targeting C#11, including suppressions for any unsupported features, if that sounds good to you

YakaryBovine:
> It would be simplest to just target C#10, but I'd also like to see the platform move forward where possible, especially when there are useful features in later versions that don't break the mapscript.
> 
> My current motivation is to bring the analyzer events down to zero. I think there were around a thousand when I checked last night.
> 
> Shall we go with option 2 and try to curate the list? Targeting C#11 initially should keep things manageable, and we can always iterate on new findings if we don't catch everything up front, or reconsider if it turns out to be too much overhead.
> 
> I'm happy to take this on and put together a draft PR for targeting C#11, including suppressions for any unsupported features, if that sounds good to you

That sounds absolutely incredible, yes please. And you're right, if we find out that the list is missing suppressions, they can easily be added later.

I would love to get the number of warnings down to zero. Handling nulls has been the biggest barrier to that for me!

PR #3445, opened 2025-09-27
feat: target C#11 with safeguards for `CSharp.lua` compatibility

alrun3:
We're using .NET 9, which defaults to C#13, but CSharp.lua only supports up to C#10. Some C#11 features (e.g., `required` properties) are already in use, but others are unsupported and can break transpilation.

To ensure compatibility and prevent invalid suggestions:

* Set `LangVersion` to 11.0 in all transpiled projects
* Add an `.editorconfig` file per transpiled project to suppress C# features not supported by CSharp.lua
* Suppress remaining informational diagnostics using `.editorconfig` and `#pragma warning disable` in generated files and vendored projects (i.e., source-copied external dependencies)

_No suppressions are added to maintained projects that fully support these features_

As a result, all informational messages have been resolved in the transpiled projects. Since the transpiled projects do not share a common directory, the `.editorconfig` settings are duplicated per project. If the directory structure is refactored in the future, these configurations could be consolidated to reduce duplication

Closes: #3440


src/WarcraftLegacies.Shared/Constants.cs
```diff
@@ -1,3 +1,4 @@
+﻿#pragma warning disable
```
YakaryBovine, line 1: This will get removed when Constants.cs and Regions.cs are regenerated, which happens on every W3X-->MapData process. I think we'll need to prepend it on generation somehow?

YakaryBovine: Cool, so sounds like we should
1. Prepend the pragma warning disable after generation, so this can be merged
2. Ask Arxos to make the same change upstream

YakaryBovine (approved):
Awesome, happy to take the patchwork Constants approach given how useful being able to just rely on IDE guidance is going to be.

alrun3:
Just repeating this here: Nice to see it merged! However, just a heads-up: the `fixup!` commits aren't meant to be included in the main branch. They're meta commits intended to be amended into their target commits. I left them as-is to make it easier to track changes since the last review. Normally, you'd either squash merge the PR or use `--autosquash` during rebase to handle these

Issue #3420, opened 2025-09-20
Remove all uses of GetHandleId native


YakaryBovine:
`GetHandleId` is a desync-prone native and has no purpose in C#/Lua. All uses of it should be removed, preferring using the object ID as a key instead. It can be found in the following files:
1. UnitEventSystem.cs
2. SpellOnAttack.cs

It is highly likely that `UnitEventSystem` can be removed entirely as it seems `PlayerUnitEvents` from `WCSharp` serves the same purpose.

PR #3446, opened 2025-09-27
refactor: replace `GetHandleId` usage with stable engine handle id

alrun3:
Replaces the usage of `GetHandleId`, which returns Lua userdata wrapper IDs, with actual engine handle IDs obtained directly from the handle objects to ensure stable identity comparisons across clients

Closes: #3420


YakaryBovine (approved):
Brilliant!

Issue #3376, opened 2025-09-09
TestMap can't be launched


YakaryBovine:
**Reproduction steps**
1. Run Launcher: TestMap: Test from your IDE
2. Observe the following runtime error

<img width="1834" height="683" alt="Image" src="https://github.com/user-attachments/assets/18909347-a0f3-4793-a42e-28185843d300" />

**Investigation so far**
We can just remove the Obsolete attributes - they're not really needed - but once you do that you get an error about the use of the Array namespace. This is very odd as the only Array function we use is also being used by WarcraftLegacies.Source. We could remove that too, but the cheat system relies heavily on it so we'd need to substitute something else.

PR #3451, opened 2025-09-28
fix: add global using for the `System` namespace

alrun3:
The TestMap has transitive dependencies that enable `ImplicitUsings`. This PR imports the `System` namespace to resolve that

Closes: #3376


YakaryBovine (approved):
Ahh beautiful, thank you

Issue #3454, opened 2025-10-01
Faction's cinematic music triggered before factions are chosen


alrun3:
In `CinematicMode`, `PlayFactionMusic` is called before `FactionChoiceDialogSetup` is shown, resulting in an attempt to play faction music before a faction is selected

PR #3455, opened 2025-10-01
fix: null deref playing faction music before choice dialog

alrun3:
Contributes to: #3454


alrun3:
> Nice, worth fixing the nullability issue but I'll keep the issue ticket open as I the intention is presumably for some music to play.

Yeah, that's why I didn't link the PR as closing the issue 👍

PR #3456, opened 2025-10-04
🚧 Run CI job

alrun3:
I am seeing a lot of failures on `dev`. Testing if it is just locally


alrun3:
Tests work when executed via `dotnet test`, however when run through Visual Studio there's a lot of:

```
------------------------ System.Collections.Generic.KeyNotFoundException : Could not find object with key '<?><?><?><?>' in the database.
---- The following constructor parameters did not have matching fixture data: MapTestFixture mapTestFixture
```

PR #3457, opened 2025-10-04
fix: restore the dummy caster `u00X`

alrun3:
Restores the dummy caster that was removed in d2f5402caab183148008365b7298c217ae3a88a3 and excludes the unit from integrity checks


YakaryBovine (approved):
Good catch, thank you!!

PR #3458, opened 2025-10-05
perf: remove `DestructibleHider` scanning logic and potential desync

alrun3:
The library had negligible performance impact in my test, introduced potential desyncs because only destructibles the player’s camera had passed would be hidden, with hidden platforms potentially causing z-level inconsistencies, and continued processing objects even after they were removed. Given the low number of destructibles present, the overhead likely outweighs any performance gain


YakaryBovine (approved):
We really should have removed this when we removed destructible trees from the game. There must be barely any destructibles left. There used to be thousands as all trees were destructibles.

PR #3469, opened 2025-10-12
perf: improve performance of unit enumeration in `LegendaryHero`

alrun3:


YakaryBovine (approved):
Woo GroupExtensions completely gone!

PR #3475, opened 2025-10-20
fix: normalize file paths for cross-platform compatibility

alrun3:
Normalizes paths based on Linux.

Windows supports both `\` and `/`, but Linux only supports `/`, so paths have been updated to use forward slashes.

### DIFF
--- a/src/Launcher/Services/W3XToMapDataConverter.cs
+++ b/src/Launcher/Services/W3XToMapDataConverter.cs
@@ -168,7 +168,7 @@ public sealed class W3XToMapDataConverter
         continue;
       }
 
-      var destinationFileName = $@"{outputFolderPath}\{ImportsPath}\{relativePath}";
+      var destinationFileName = Path.Combine(outputFolderPath, ImportsPath, relativePath);
       Directory.CreateDirectory(Path.GetDirectoryName(destinationFileName)!);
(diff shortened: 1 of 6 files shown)


src/Launcher/Services/W3XToMapDataConverter.cs
```diff
@@ -178,34 +178,27 @@ private static void CopyUnserializableFiles(string baseMapPath, string outputFol
...
-      var fullOutputFilePath = $@"{outputFolderPath}\{filePath}";
-      if (fullSourceFilePath.EndsWith(".txt"))
-      {
-        //Warcraft 3 maps use CR line endings, so this ensures that those line endings are replaced with CLRF.
-        File.WriteAllLines(fullOutputFilePath, File.ReadLines(fullSourceFilePath));
-      }
-      else
-      {
-        File.Copy(fullSourceFilePath, fullOutputFilePath, true);
-      }
```
alrun3: This did not enforce `CRLF` and I had no problems running without it. Let me know if `CRLF` enforcement needs to be added

alrun3:
Ready for review

YakaryBovine (approved):
Looks good, and I'll keep Linux compatibility in mind going forward.

PR #3478, opened 2025-10-21
Revert erroneous removal of the Evasion component of Akama's Shadow Aura ability

YakaryBovine:
In https://github.com/AzerothWarsLR/WarcraftLegacies/commit/8aecac493315fa48d71bbbc96f69fb80b3e6a3b4, the registration for Akama's Shadow Aura ability was erroneously removed, and the associated object data and underlying `Spell` were removed subsequently. This PR puts them all back, restoring Shadow Aura to working form, and refactors Shadow Aura to no longer conduct its registration logic straight in the constructor.


src/WarcraftLegacies.Shared/Constants.cs
```diff
@@ -8,7 +8,7 @@
 /// </summary>
 public static class Constants
 {
-	public const int ABILITY_ANSL_SLEEP_NZOTH = 1095652172;
```
alrun3: Have these always been unstable? I think we should add some sorting to prevent them from changing position

alrun3: Created: https://github.com/AzerothWarsLR/WarcraftLegacies/issues/3479

YakaryBovine:
I had to redo the refactor because it turns out IEffectOnLearn is a Spell feature, not a PassiveAbility one.

src/WarcraftLegacies.Source/PassiveAbilities/AddAbilityOnLearn.cs
```diff
@@ -0,0 +1,32 @@
...
+    var targetLevel = learner.GetAbilityLevel(Id);
+
+    if (learner.GetAbilityLevel(AbilityToAddId) == 0)
+    {
+      learner.AddAbility(AbilityToAddId);
+    }
+
+    learner.SetAbilityLevel(AbilityToAddId, targetLevel);
+    Console.WriteLine(targetLevel);
+    Console.WriteLine(learner.GetAbilityLevel(AbilityToAddId));
```
alrun3: Yeet

PR #3488, opened 2025-10-22
Move Power, Hazard, Channel, and Missile implementations out of MacroTools

YakaryBovine:


src/MacroTools/Channels/Channel.cs
```diff
@@ -1,7 +1,7 @@
 ﻿using System;
 using WCSharp.Events;
 
-namespace MacroTools.ChannelSystem;
+namespace MacroTools.Channels;
```
alrun3: To me, this name denotes either a communication stream or a waterway. Is channeling separate from abilities or could it be merged into the spell namespace?

src/WarcraftLegacies.Source/Powers/DummyPower.cs
```diff
@@ -0,0 +1,21 @@
+﻿using MacroTools.FactionSystem;
+
+namespace WarcraftLegacies.Source.Powers;
+
+public sealed class DummyPower : Power
```
alrun3: What do we need a `DummyPower` for in the map script? If its purpose is to create an instantiable power, why isn't `Power` made instantiable?

src/MacroTools/Channels/Channel.cs
YakaryBovine: "Channel" is WC3 & WoW terminology; a casted ability that has to be continuously maintained to keep doing its effect.

Yeah... Each Channel is very much tied to an ability ID at the moment so it could quite possibly belong in that namespace. I'll have a think about it, there might be a use case for a non-spell channel. Though probably not.

src/WarcraftLegacies.Source/Powers/DummyPower.cs
YakaryBovine: The former, because there are Powers that render in the UI, but their underlying effect is handled by something else. So they're dummies.

`Power` could be made instantiable but I tend towards the idea that any given class should be either `abstract` or `sealed`.  `Power` is saying "implement me", `DummyPower` is just the saddest possible implementation. Up for debate though no doubt.

PR #3490, opened 2025-10-22
Use inverted endianness in UnitTooltipMigration

YakaryBovine:
https://github.com/AzerothWarsLR/WarcraftLegacies/commit/9adb048757e1b98ac1cc73b58af80ff3a4fe1ac1 changed ObjectInfoRepository to use numeric values, which for whatever reason have inverted endianness from what gets extracted by War3Net, causing the UnitTooltipMigration to break. This inverts the endianness on the UnitTooltipMigration end.

### DIFF
--- a/src/Launcher/MapMigrations/UnitTooltipMigration.cs
+++ b/src/Launcher/MapMigrations/UnitTooltipMigration.cs
@@ -4,6 +4,7 @@ using System.Text;
 using Launcher.Extensions;
 using War3Api.Object;
 using War3Net.Build;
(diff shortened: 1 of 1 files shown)


alrun3 (approved):
Ah yes, an `int` is an `int`, but a `FourCC` is not a `RawCode` (they differ in endianness). The former is used in the game, and the latter in the map file. I haven't investigated the reason for the discrepancy, and this will fix it, but we should ask ourselves why we are comparing RawCodes to FourCCs in the first place? And preferably address the root cause

Issue #3411, opened 2025-09-17
Illidari tooltip review


YakaryBovine:

YakaryBovine:
- All units are missing their limits!
- Abilities that can't be leveled up despite having [Level 1] in name:
   - Forked Lightning
   - Ensnare
   - War Plating
- Redundant hotkey indicators:
   - Cold Arrows
   - Devour
- Remove "Ashtongue" from unit names
- Redo unit flavour to describe unit function:
   - Royal Guard
   - Sea Witch
   - Siren
   - Coilfang Summoner
   - Couatl
   - Scylla
   - Myrmidon
   - Snap Dragon
   - Dragon Turtle
   - Deathsworn
   - Mystic
- Remove dashes from train and build tooltip
- Fix lumber cost on Najentus (doesn't appear to have worked)
- Mur'gul Shadowcaster says "Attacks land and air units." twice
- Coilfang Summoner's Tidal Serpent does nothing (probably removed by accident)
- Monsoon and Forked Lightning share an icon
- Vindication has inappropriate flavour; Draenei don't use the Light
- Improved Tidal Watcher requires Navigation
- Vashj's Tornado is called Hurricane for no reason... It says it deals damage to units but it doesn't
- Vashj's Mana Shield is missing a space
- Shadow Aura debuff is green and refers to "fear aura"
- Vanish briefly applies a Sam buff
- Summon Mur'gul tooltip just sucks, and buff is Feral Spirit

PR #3497, opened 2025-10-24
Add tooltip migration for Tooltip - Basic and Tooltip - Revive fields

YakaryBovine:
Adds a migration that automatically determines the Tooltip - Basic, Tooltip - Awaken, and Tooltip - Revive fields for all units.
Contributes to #3411 by fixing the issue mentioned where some units trained have an erroneous dash in their tooltip.

### DIFF
--- a/src/Launcher/MapMigrations/MapMigrationProvider.cs
+++ b/src/Launcher/MapMigrations/MapMigrationProvider.cs
@@ -10,7 +10,7 @@ public static class MapMigrationProvider
       new CreepLevelMapMigration(),
       new GoldBountyMapMigration(),
       new FlightMigration(),
-      new UnitTooltipMigration(),
+      new UnitTooltipExtendedMigration(),
       new UnitTooltipBasicMigration(),
       new PortraitModelFileMapMigration()
     };
--- a/src/Launcher/MapMigrations/UnitTooltipExtendedMigration.cs
+++ b/src/Launcher/MapMigrations/UnitTooltipExtendedMigration.cs
@@ -12,7 +12,7 @@ namespace Launcher.MapMigrations;
 /// <summary>
 /// Sets all unit tooltips in the game.
 /// </summary>
-public sealed class UnitTooltipMigration : IMapMigration
+public sealed class UnitTooltipExtendedMigration : IMapMigration
 {
   private const string LineSeparator = "|n";
   private const string AbilitiesKnown = "|cfff5962dAbilities:|r ";


src/Launcher/MapMigrations/UnitTooltipBasicMigration.cs
```diff
@@ -0,0 +1,64 @@
...
+/// <summary>
+/// Sets the "Text - Tooltip - Basic", "Text - Tooltip - Revive", and "Text - Tooltip - Awaken" fields for units.
+/// </summary>
+public sealed class UnitTooltipBasicMigration : IMapMigration
+{
+  /// <inheritdoc />
+  public void Migrate(Map map, ObjectDatabase objectDatabase)
+  {
+    var units = objectDatabase.GetUnits();
+    var copiedUnits = units.ToList();
```
alrun3: Is the intention here to materialize a LINQ sequence, or is it meant to create independent copies of the units?
```suggestion
    var units = objectDatabase.GetUnits().ToList();
```

src/Launcher/MapMigrations/UnitTooltipBasicMigration.cs
```diff
@@ -0,0 +1,64 @@
...
+      trainType = "Train";
+    }
+
+    return trainType;
+  }
+
+  /// <summary>
+  /// Gets the best name for a unit, preferring hero proper name over their text name.
+  /// </summary>
+  private static string GetBestName(Unit unit) => unit.TextProperNames.Any() ? unit.TextProperNames.First() : unit.TextName;
```
alrun3: Can we always count on there being only a single unit for us to rely on the first TextProperName?

alrun3 (approved):

src/Launcher/MapMigrations/UnitTooltipBasicMigration.cs
YakaryBovine: In Legacies, yes; all heroes have exactly 1 Proper Name, since they represent specific characters in the Warcraft canon.

src/Launcher/MapMigrations/UnitTooltipBasicMigration.cs
```diff
@@ -0,0 +1,64 @@
...
+        Console.WriteLine($"Failed to apply tooltip migration for {unit.TextName}: {ex}");
+      }
+    }
+
+    var unitData = objectDatabase.GetAllData().UnitData;
+    map.UnitObjectData = unitData;
+    map.UnitSkinObjectData = unitData;
+  }
+
+  private static string GetTrainType(Unit unit)
```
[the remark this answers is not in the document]
YakaryBovine: Agreed.

src/Launcher/MapMigrations/UnitTooltipBasicMigration.cs
YakaryBovine: Accepted the suggestion.
Yeah, it's to create independent copies of the units. Drake53 explained to me why in his Hive Workshop post about C# but I didn't entirely understand - it just doesn't work without the `ToList`.

src/Launcher/MapMigrations/UnitTooltipBasicMigration.cs
```diff
@@ -0,0 +1,64 @@
...
+          unit.TextTooltipRevive = $"Revive {unit.TextName}";
+        }
+      }
+      catch (Exception ex)
+      {
+        Console.WriteLine($"Failed to apply tooltip migration for {unit.TextName}: {ex}");
+      }
+    }
+
+    var unitData = objectDatabase.GetAllData().UnitData;
```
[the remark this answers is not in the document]
YakaryBovine: It's different actually. The first statement extracts the data as a list of War3Api.Object `Unit`s, which we then modify in place. The second statement returns War3Net `UnitObjectData` from `War3Api.Object` `ObjectDatabase`, which is what we need to supply to `War3Net.Map`'s `UnitObjectData`.

src/Launcher/MapMigrations/UnitTooltipBasicMigration.cs
YakaryBovine: I tried getting rid of the `ToList` and it worked just fine so... I don't know. Removed it, no more materialization.

PR #3499, opened 2025-10-25
Fix level 3 Exact Justice dealing no damage

### LINKED ISSUE #3228: Uther's Ultimate ability not dealing damage on Rank 3
from a community report need to check

YakaryBovine:
Fixes an issue where the Channel class was determining duration based off a 1-indexed ability field which is actually 0-indexed. Incidentally fixed the damage values in the Exact Justice tooltip.

Closes #3228

### DIFF
--- a/mapdata/WarcraftLegacies/AbilityData/Skin/926494785.json
+++ b/mapdata/WarcraftLegacies/AbilityData/Skin/926494785.json
@@ -36,7 +36,7 @@
     {
       "Id": 1718840929,
       "Type": 3,
-      "Value": "(Purple - Uther)"
+      "Value": "(Uther)"
     },
     {
       "Id": 1952540001,
@@ -75,7 +75,7 @@
     {
       "Id": 1953854049,
       "Type": 3,
-      "Value": "The hero channels the Light for 6 seconds, preventing all nearby friendly units from taking damage. When Divine Justice finishes casting, all nearby enemy units take damage. If Divine Justice is interrupted, it instead deals damage proportional to how long it was channeled for.|n|n|cffffcc00Level 1|r - Deals 150 damage. |n|cffffcc00Level 2|r - Deals 350 damage. |n|cffffcc00Level 3|r - Deals 600 damage."
+      "Value": "The hero channels the Light for \u003CA097,Dur1\u003E seconds, preventing all nearby friendly units from taking damage. When Divine Justice finishes casting, all nearby enemy units take damage. If Divine Justice is interrupted, it instead deals damage proportional to how long it was channeled for.|n|n|cffffcc00Level 1|r - Deals 200 damage. |n|cffffcc00Level 2|r - Deals 400 damage. |n|cffffcc00Level 3|r - Deals 600 damage."
     },
     {
       "Id": 1952543841,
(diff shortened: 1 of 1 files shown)


src/MacroTools/Channels/Channel.cs
```diff
@@ -19,13 +19,12 @@ protected Channel(unit caster, int spellId)
     Caster = caster;
     SpellId = spellId;
     var ability = Caster.GetAbility(SpellId);
-    var durationFromObjectEditor = BlzGetAbilityRealLevelField(ability, ABILITY_RLF_DURATION_NORMAL,
-      Caster.GetAbilityLevel(SpellId));
+    var abilityLevel = Caster.GetAbilityLevel(SpellId);
+    var durationFromObjectEditor = ability.GetDurationNormal_adur(abilityLevel - 1);
```
alrun3: nit: Could we extract `abilityLevel -1` to something like `levelIndex`?

src/MacroTools/Channels/Channel.cs
```diff
@@ -19,13 +19,12 @@ protected Channel(unit caster, int spellId)
     Caster = caster;
     SpellId = spellId;
     var ability = Caster.GetAbility(SpellId);
-    var durationFromObjectEditor = BlzGetAbilityRealLevelField(ability, ABILITY_RLF_DURATION_NORMAL,
-      Caster.GetAbilityLevel(SpellId));
+    var abilityLevel = Caster.GetAbilityLevel(SpellId);
```
alrun3: There's potentially still a need to ensure the level index lies within the closed interval of `[1, ABILITY_IF_LEVELS]`

YakaryBovine: I think that's logically guaranteed, right? We're only checking fields based on the actual level of the ability, so the field must necessarily be available.

mapdata/WarcraftLegacies/AbilityData/Skin/926494785.json
```diff
@@ -75,7 +75,7 @@
     {
       "Id": 1953854049,
       "Type": 3,
-      "Value": "The hero channels the Light for 6 seconds, preventing all nearby friendly units from taking damage. When Divine Justice finishes casting, all nearby enemy units take damage. If Divine Justice is interrupted, it instead deals damage proportional to how long it was channeled for.|n|n|cffffcc00Level 1|r - Deals 150 damage. |n|cffffcc00Level 2|r - Deals 350 damage. |n|cffffcc00Level 3|r - Deals 600 damage."
+      "Value": "The hero channels the Light for \u003CA097,Dur1\u003E seconds, preventing all nearby friendly units from taking damage. When Divine Justice finishes casting, all nearby enemy units take damage. If Divine Justice is interrupted, it instead deals damage proportional to how long it was channeled for.|n|n|cffffcc00Level 1|r - Deals 200 damage. |n|cffffcc00Level 2|r - Deals 400 damage. |n|cffffcc00Level 3|r - Deals 600 damage."
```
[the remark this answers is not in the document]
YakaryBovine: It would be _extremely_ nice, but I think that's an enormous piece of work. We've just been tracking those in an "epics" channel in Discord, so as to reserve Issues for things that will be done sooner rather than later.

PR #3505, opened 2025-10-25
Move Titanforging mechanic from MacroTools to WarcraftLegacies.Source

YakaryBovine:
The main purpose of this is actually to support the InaccessibleObjectCollection, which does not account for ability IDs only present in MacroTools. Rather than make it consider MacroTools, I'd rather just not mention Legacies-specific object IDs in there at all.


src/WarcraftLegacies.Source/Spells/TitanforgeArtifact.cs
```diff
@@ -26,24 +34,37 @@ public override void OnCast(unit caster, unit target, Point targetPoint)
       var heldItem = @event.Unit.ItemAtOrDefault(0);
```
alrun3: Is it known that the artifact must be in the first slot?

Issue #3392, opened 2025-09-13
Upgrading Hardened after Spell Conduction changes Eradicator armor type to Light


YakaryBovine:
1. Queue up the Spell Conduction research _then_ Shaped Obsidian
2. Once Shaped Obsidian completes, observe that there is a refund message, but the armor type of Eradicators is changed to Light

Note that Shaped Obsidian is meant to change armor type to Heavy, and when it gets refunded it should not be changed at all.

Arkaham:
Hi, I was trying to work on this issue for a bit and couldn't fix it, as I'm interested in the project and would like to help, could you please give me a tip, i was trying to both work with FactionManager to prevent being able to queue 2 Incompatible researches and work with json files and their values but neither worked, am I looking at the wrong place ?

YakaryBovine:
> Hi, I was trying to work on this issue for a bit and couldn't fix it, as I'm interested in the project and would like to help, could you please give me a tip, i was trying to both work with FactionManager to prevent being able to queue 2 Incompatible researches and work with json files and their values but neither worked, am I looking at the wrong place ?

It's a pretty hard problem and may ultimately be down to a Warcraft 3 bug preventing the research from being properly reverted. The bug didn't exist until a WC3 patch came out. Your investigations sound pretty reasonable to me.

We might just have to change the implementation of the upgrades to do something other than change Armor Type.

You're welcome to join our dev Discord if you want to chat: https://discord.gg/zZyTENpA

PR #3516, opened 2025-11-25
Redesigning of Eradicator researches.

Arkaham:
As it was said in [#3392](https://github.com/AzerothWarsLR/WarcraftLegacies/issues/3392), there is a bug regarding upgrade of an armor to heavy, if refunded, the armor would go to light instead of medium, after discussion we came to conclusion that the redesign of those researches is needed.

Can you please check if the both researches are correct ? I'm open for a discussion.

### DIFF
--- a/src/WarcraftLegacies.Source/UnitTypeTraits/SpellConduction/SpellConductionBuff.cs
+++ b/src/WarcraftLegacies.Source/UnitTypeTraits/SpellConduction/SpellConductionBuff.cs
@@ -1,5 +1,6 @@
 ﻿using System.Linq;
 using MacroTools.Extensions;
+using MacroTools.UnitTypeTraits;
 using WCSharp.Buffs;
 using WCSharp.Events;
 
@@ -44,14 +45,13 @@ public sealed class SpellConductionBuff : PassiveBuff
       return;
     }
 
-    var attackType = @event.AttackType;
-    if (!IsRedirectableAttackType(attackType))
+    if (UnitTypeTraitRegistry.UnitHasTrait(Target, typeof(SpellConductionTrait)))
     {
       return;
     }
 
-    var damageSource = @event.DamageSource;
-    if (damageSource != null && HasSpellConduction(damageSource))
+    var attackType = @event.AttackType;
+    if (!IsRedirectableAttackType(attackType))
     {
       return;
     }
@@ -67,7 +67,4 @@ public sealed class SpellConductionBuff : PassiveBuff
       @event.DamageType, @event.WeaponType);
 
   private bool IsRedirectableAttackType(attacktype attackType) => RedirectableAttackTypes.Contains(attackType);
-
-  private static bool HasSpellConduction(unit unit) =>
-    BuffSystem.GetBuffsOnUnit(unit).OfType<SpellConductionBuff>().Any();
 }
--- a/src/MacroTools/UnitTypeTraits/UnitTypeTraitRegistry.cs
+++ b/src/MacroTools/UnitTypeTraits/UnitTypeTraitRegistry.cs
@@ -1,4 +1,5 @@
-﻿using System.Collections.Generic;
+﻿using System;
+using System.Collections.Generic;
 using MacroTools.Utils;
 using WCSharp.Events;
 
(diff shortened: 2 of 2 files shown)


mapdata/WarcraftLegacies/UpgradeData/ZBHS.json
```diff
@@ -102,11 +77,11 @@
       "Level": 1,
       "Id": 828536167,
       "Type": 3,
-      "Value": "Changes Obsidian Eradicators\u0027 armor type to Heavy and replaces their Spell Resistance with Hardened, which causes them to take \u003CZBHL,DataC1\u003E reduced damage from attacks.|n|n|cffc864c8Can\u0027t be researched with Spell Conduction.|r"
+      "Value": "Increases Obsidian Eradicators maximum hit points by 200.|n|n|cffc864c8Can\u0027t be researched with Spell Conduction.|r"
```
YakaryBovine: There's no need to get rid of the Hardened ability, I think. We just want to remove the Heavy armor conversion.

src/WarcraftLegacies.Source/UnitTypeTraits/SpellConduction/SpellConductionBuff.cs
```diff
@@ -61,4 +72,7 @@ private void DamageCaster(unit damager, float eventDamage) =>
       damagetype.Universal, @event.WeaponType);
```
YakaryBovine: One of the things we discussed wanting to do was making Spell Conduction correctly take into account the resistances of the Eradicator. The reason it doesn't right now is this `damagetype.Universal`. If we change that to the damage type of the original damage, it will take the Eradicator's resistances into account.

mapdata/WarcraftLegacies/UpgradeData/ZBML.json
```diff
@@ -102,7 +102,7 @@
       "Level": 1,
       "Id": 828536167,
       "Type": 3,
-      "Value": "Allows Obsidian Eradicators to use Spell Conduction, which causes 35% of Magic and Spell damage dealt to friendly units within a radius of 500 to be redirected to them. The damage is affected by the resistances of both the damaged unit and the Obsidian Eradicator. Also grants 100 extra Hit Points |n|n|cffc864c8Can\u0027t be researched with Shaped Obsidian.|r"
+      "Value": "Allows Obsidian Eradicators to use Spell Conduction, which causes 40% of Magic and Spell damage dealt to friendly units within a radius of 500 to be redirected to them. Damage from other Spell Conduction units is not redirected. The damage is affected by the resistances of both the damaged unit and the Obsidian Eradicator. Also grants 100 extra Hit Points.|n|n|cffc864c8Can't be researched with Shaped Obsidian.|r"
```
YakaryBovine: I think it's not necessary to tell the player that Spell Conduction doesn't bounce; logically it can't, since that would be an infinite loop.
```suggestion
      "Value": "Allows Obsidian Eradicators to use Spell Conduction, which causes 40% of Magic and Spell damage dealt to friendly units within a radius of 500 to be redirected to them. The damage is affected by the resistances of both the damaged unit and the Obsidian Eradicator.|n|n|cffc864c8Can't be researched with Shaped Obsidian.|r"
```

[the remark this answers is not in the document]
Arkaham: The extra 100 hit points were removed as we discussed.

src/WarcraftLegacies.Source/UnitTypeTraits/SpellConduction/SpellConductionBuff.cs
```diff
@@ -20,6 +20,11 @@ public sealed class SpellConductionBuff : PassiveBuff
   /// </summary>
   public required attacktype[] RedirectableAttackTypes { get; init; }
 
+  /// <summary>
+  /// The ability ID to check for Spell Conduction.
+  /// </summary>
+  private const int SpellConductionAbilityId = 1380794970;
```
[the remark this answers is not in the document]
Arkaham: The buff code was removed, i forgot it is actually an ID from editor not the game itself, can you please check if what i changed is how it is supposed to look?

mapdata/WarcraftLegacies/AbilityData/ZBMR.json
```diff
@@ -35,7 +35,7 @@
       "Level": 1,
       "Id": 828536161,
       "Type": 3,
-      "Value": "Causes 35% of Magic and Spell damage dealt to friendly units within a radius of 500 to be redirected to this unit. The damage is affected by the resistances of both the damaged unit and this unit."
+      "Value": "Causes 40% of Magic and Spell damage dealt to friendly units within a radius of 500 to be redirected to this unit. Damage from other Spell Conduction units is not redirected. The damage is affected by the resistances of both the damaged unit and this unit."
```
YakaryBovine:
```suggestion
      "Value": "Causes 40% of Magic and Spell damage dealt to friendly units within a radius of 500 to be redirected to this unit. The damage is affected by the resistances of both the damaged unit and this unit."
```

YakaryBovine (approved):
Nice work!!

PR #3519, opened 2025-12-13
Optimize speed of test command by producing a folder rather than a file

YakaryBovine:
The Test profile takes a very long time to run, in the order of 12 seconds (see SaveMapFile below).
image

It turns out that the bulk of this time is spent producing the MPQ archive. Switching the test mode to build a directory instead (see BuildDirectory below) drastically reduces this time.
image

While this speed test is not exactly scientific, it's an 88% reduction in time so I'm happy to call Rider's performance output good enough.

### DIFF
--- a/src/Launcher/PathConventions.cs
+++ b/src/Launcher/PathConventions.cs
@@ -4,8 +4,6 @@ namespace Launcher;
 
 public static class PathConventions
 {
-  public const string SharedImports = "shared_imports";
-
   public const string Maps = "maps";
 
   public const string MapData = "mapdata";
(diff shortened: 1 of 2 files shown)


alrun3 (approved):
There's some more low-hanging fruit we could consider. What the test profile effectively does is apply migrations, change the starting player, and launch the map. As such, if only code changes were made, you should use `csharp-to-lua`, as it already operates on folders rather than archives. If only changes to, e.g., terrain were made, you should just launch the map.

If you've made changes that require migrations, it'd be nice if those could be made updatable, so you don't have to destroy the map; basically creating an incremental testing strategy, with the optional possibility of doing a full rebuild to save even more time

PR #3533, opened 2025-12-24
Generate Regions as properties rather than fields to avoid local variable limit

YakaryBovine:


src/Launcher/Services/ConstantsGenerator.cs
```diff
@@ -76,10 +76,10 @@ private void GenerateRegions(List<Region> regions)
 
     foreach (var region in regions)
     {
-      var fieldName = Sanitize(region.Name.Replace(" ", "_"));
-      var fieldValue = string.Create(System.Globalization.CultureInfo.InvariantCulture,
+      var propertyName = Sanitize(region.Name.Replace(" ", "_"));
+      var propertyValue = string.Create(System.Globalization.CultureInfo.InvariantCulture,
         $"new Rectangle({region.Left}f, {region.Bottom}f, {region.Right}f, {region.Top}f)");
-      writer.WriteLine($"  public static readonly Rectangle {fieldName} = {fieldValue};");
+      writer.WriteLine($"  public static Rectangle {propertyName} {{ get; set; }} = {propertyValue};");
```
alrun3: Is the setter needed?
```suggestion
      writer.WriteLine($"  public static Rectangle {propertyName} {{ get; }} = {propertyValue};");
```

YakaryBovine: Oddly, yes I think so - without them these get inlined into a function and we hit the local variable limit.

PR #3538, opened 2025-12-25
chore: remove deprecated map trigger support

alrun3:


YakaryBovine (approved):
Hah nice catch. Yeah, at some point we did have a hybrid of GUI triggers and C#, when we were still transitioning to code-only.

PR #3541, opened 2025-12-26
Replace AutoMapper with War3Net's built-in JsonConverters

alrun3:
This PR demonstrates the impact of removing DTOs and relying solely on War3Net for serialization.

Benefits
* Significantly less C# code to maintain
* Removes duplication between DTOs and War3Net models
* Improved compatibility with current and future file format features

Trade-off
* Map data files become larger due to storing all fields, including default values


YakaryBovine (approved):
Looks amazing. Nice one-million lines of changes!

Issue #3513, opened 2025-10-28
Make MapData<-->W3X conversion compatible with HiveWE


YakaryBovine:
The HiveWE is an incredibly fast editor and we'd like to be able to use it for object editing and terraining. We can't use it right now because it stores data in a different format than the World Editor, and we want to maintain compatibility with both. In particular:

**Incompatibilities**
1. It doesn't split object data into Core and Skin files, like the World Editor does.
2. It doesn't automatically generate a pathing map on save.
3. It doesn't automatically generate unit Z positions on save.
4. It seems to have a higher precision for unit X/Y positions.

**Compatibility goals**
So, we have 2 key compatibility goals:
1. A .w3x created from MapData can be opened with both HiveWE and World Editor.
2. A map saved in HiveWE and a map saved in World Editor produce the _same_ MapData.

**Compatibility implementation**
To achieve compatibility, we must:

- [x] Store object data in a single file per object, rather than splitting between Core and Skin. For maps saved in the World Editor, this means explicitly merging objects.
- [x] Replace the WCSharp ConstantGenerator with our own implementation that doesn't rely on TriggerStrings.
- [x] Put Z position saving for preplaced units back in (HiveWE won't render units properly without it).
- [x] Reduce the precision of X and Y unit positions to ensure HiveWE and WorldEditor produce the same result.

PR #3543, opened 2025-12-27
Re-export MapData from HiveWE then World Editor

YakaryBovine:
**Background**
As alluded to in #3513, saving the map in World Editor seems to produce different unit and doodad positions than saving the map in HiveWE. While investigating this, I stumbled upon two facts:

1. Re-saving WarcraftLegacies in HiveWE produces trivial changes to X/Y positions, but resaving it in World Editor does **not** further change these positions.
2. Subsequently re-saving WarcraftLegacies in World Editor produces trivial changes to Z positions based on recalculations resulting from (1), but then resaving in HiveWE does **not** further change those positions.

**Change**
Thus, it is possible to produce a stable set of unit positions by first re-saving in HiveWE and then re-saving in World Editor. This is what this PR does.

With this merged we would be able to use HiveWE as a tool for both object editing and unit placement.

**Caveats**
This is not a lasting solution, however; if a developer were to place or move a unit in World Editor, they would subsequently experience the same churn to reach stable data as per (1) and (2). That is quite acceptable in the short term, since the intent is for all developers to use HiveWE as their primary editor, and this doesn't lock us into anything.


YakaryBovine:
I'm marking this as a draft because it's probably going to be more practical to do it after #3541, otherwise Alrun will have to repeat his exports.

alrun3:
> I'm marking this as a draft because it's probably going to be more practical to do it after https://github.com/AzerothWarsLR/WarcraftLegacies/pull/3541

No worries. This is good to merge provided it yields stable position data across all three scenarios:

1. Creating a map from mapdata and exporting it back to mapdata
2. Creating a map from mapdata, opening it in HiveWE, saving, and exporting it to mapdata
3. Creating a map from mapdata, opening it in World Editor, saving, and exporting it to mapdata

It is unfortunate that HiveWE produces slight discrepancies in position data. However, we can always restore full precision from version control if it converges in the future.

> Also, it's very possible that that work will change how this works anyway.

The only thing that comes to mind is that I'd really like to see Z-position data added to units

PR #3544, opened 2025-12-27
fix: replace `WCSharp` fork with upstream `War3Net`

alrun3:
This PR switches back to upstream War3Net, as the WCSharp fork contains a serialization bug where `FormatVersion` is not propagated correctly, causing incorrect rules to be applied

Supersedes: https://github.com/AzerothWarsLR/WarcraftLegacies/pull/3375
Contributes to: https://github.com/AzerothWarsLR/WarcraftLegacies/pull/3541


YakaryBovine (approved):
LOL right, so that's why I wrote that weird TerrainTileResolver...

PR #3545, opened 2025-12-27
fix: rename `ANst` to `ANST` to match rawcode

alrun3:
Causes a diff on case-sensitive operating systems as `1414745665` is `ANST`.

Fun-fact: `ANst` is stampede. If that base ability had been modified, it would have caused a path collision on case-insensitive operating systems due to using rawcodes as filenames


YakaryBovine (approved):
That's the second one of these you've found, I wonder how they happened in the first place?

PR #3546, opened 2025-12-28
Support W3XToMapData incremental exports

YakaryBovine:
This PR adds support for doing incremental MapData exports. W3XToMapDataConverter can take in a new `IncludeInMap` flag, which is used to:
- Determine which paths to delete prior to performing an export
- Indicate to `Map.Open` which files it should absorb from the provided W3X file

The `w3x-to-json serialize` command has been updated to support passing through these flags, defaulting to `All`. There are are two utility flags, `ObjectData` and `Terrain`, that encompass a number of other flags. Finally, a new default has been added to `launchSettings.json` for exporting only object data, since that is a very common editor need.

These changes will help speed up development by making exports faster, as long as the developer is willing to use the command line and has an understanding of what they need to export.

### DIFF
--- a/src/Launcher/CLI/Commands/MapCommand.cs
+++ b/src/Launcher/CLI/Commands/MapCommand.cs
@@ -1,6 +1,7 @@
 ﻿using System;
 using System.CommandLine;
 using Launcher.CLI.Contexts;
+using Launcher.Services;
 
 namespace Launcher.CLI.Commands;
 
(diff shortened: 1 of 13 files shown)


alrun3 (approved):
nit: This adds support for partial exports, but still deletes the destination. For incremental exports, we need to be able to avoid deleting the destination, e.g., by adding a configurable delete behavior that can be disabled

YakaryBovine:
> nit: This adds support for partial exports, but still deletes the destination. For incremental exports, we need to be able to avoid deleting the destination, e.g., by adding a configurable delete behavior that can be disabled

As discussed on Discord #programming, I'll add a configuration option to not delete destination folders/files.

PR #3548, opened 2025-12-28
chore: upgrade target framework to net10.0

alrun3:


alrun3:
War3Net requires all build parameters to be set via the csproj

PR #3553, opened 2025-12-29
ci: bump `actions/checkout` to v6

alrun3:
`actions/checkout@v3` is deprecated and uses a deprecated NodeJS runtime


YakaryBovine (approved):
Wow we were very behind!

PR #3556, opened 2025-12-30
feat: add system for retrieving preplaced widgets

alrun3:
Saw the `PreplacedUnitSystem` being passed explicitly between classes, so I made it static to simplify access and added support for all widget types (units, destructables, items).

Opening this in case you want to review or merge. Full disclaimer: I haven’t had a chance to test it yet.

Side-note: We should consider moving all systems to the `MacroTools.Systems` namespace regardless of their solution structure for additional simplicity


YakaryBovine (approved):
Awesome stuff code-wise, I just need a moment to test it before I merge.
I think this would be an obvious case for a library that would be fantastic for other macro maps to consume, as interacting with preplaced units is an extremely common need in the genre.

PR #3559, opened 2026-01-01
Support MapDataToW3X incremental exports

YakaryBovine:
In the opposite direction to #3546, this PR allows developers to convert only parts of MapData and transfer them to the output `maps` directory. This defaults to exporting everything _except_ the script file, since the default conversion exists to produce a map for editing in the World Editor. I have also included a new launch setting to demonstrate exporting the script file by itself.

### DIFF
--- a/src/Warcraft.Cartographer/Deserialization/MapDataToMapConverter.cs
+++ b/src/Warcraft.Cartographer/Deserialization/MapDataToMapConverter.cs
@@ -64,8 +64,13 @@ public sealed class MapDataToMapConverter(MapDataToMapConverterOptions options)
     return map;
   }
 
-  private MapDoodads DeserializeDoodads()
+  private MapDoodads? DeserializeDoodads()
   {
+    if (!options.IncludeFromMap.HasFlag(IncludeFromMap.Doodads))
+    {
+      return null;
+    }
+
     var mapDoodads = new MapDoodads(MapWidgetsFormatVersion.v8, MapWidgetsSubVersion.v11, true);
     foreach (var file in Directory.EnumerateFiles(options.MapDataPaths.DoodadsPath))
     {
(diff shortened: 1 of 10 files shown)


src/Warcraft.Cartographer/Deserialization/MapDataToMapConverter.cs
```diff
@@ -275,6 +300,11 @@ private UnitObjectData DeserializeUnitData()
 
   private List<PathData> GetAdditionalFiles()
   {
+    if (!options.IncludeFromMap.HasFlag(IncludeFromMap.Imports))
+    {
+      return [];
+    }
```
alrun3: Doesn't this method also include the gameplay constants, minimap and game interface?

YakaryBovine: My bad, thank you for the spot. Fixed to check each flag individually in both methods.

src/Warcraft.Cartographer/Deserialization/MapDataToMapConverter.cs
```diff
@@ -305,6 +335,11 @@ private List<PathData> GetAdditionalFiles()
 
   private List<DirectoryEnumerationOptions> GetAdditionalFileDirectories()
   {
+    if (!options.IncludeFromMap.HasFlag(IncludeFromMap.Imports))
+    {
+      return [];
+    }
```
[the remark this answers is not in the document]
YakaryBovine: Yeah, maybe. It might be possible to just only produce the directory enumeration options? The reason it's like this is because the MPQArchive from War3Net needs the input in a directory/search pattern format, whereas for our custom directory writing purposes it copies the individual files. I have no idea why I did it that way.

Copying the entire folder sounds great but... Does .NET even provide a way to copy entire folders? I thought all copies were file-by-file.

I agree this is worth improving for both complexity and performance reasons, but for the present I'm keen to move on to gameplay content so I will merge this as-is. We could follow-up chat on Discord.

PR #3561, opened 2026-01-02
Integrate unit categories into automatically generated unit tooltips

YakaryBovine:
Contributes to #3411 

Unit tooltips in Legacies are a bit of a mess, often focusing entirely on flavour without conveying any. For instance, the tooltip for "Blood Mage" is "A mage who has given into the lure of blood magic." This is pointless - the flavour specified is already well explained by the unit name.

So instead, we're applying designation categories to every unit, and using those to form the first line of the unit tooltip. This should provide quick information for the player without burdening us with writing tonnes of tooltips, and will hopefully provide some design direction on those units. Examples below; see the teal text.

image

image

### DIFF
--- a/src/MacroTools.Shared/UnitCategory.cs
+++ b/src/MacroTools.Shared/UnitCategory.cs
@@ -34,7 +34,8 @@ public enum UnitCategory
   Siege,
   Destroyer,
   Tank,
-  Assassin
+  Assassin,
+  Summoner
 }
 
 public static class UnitCategoryExtensions
@@ -95,6 +96,8 @@ public static class UnitCategoryExtensions
         return "Tank";
       case UnitCategory.Assassin:
         return "Assassin";
+      case UnitCategory.Summoner:
+        return "Summoner";
     }
 
     return category.ToString();
--- a/src/WarcraftLegacies.Shared/FactionObjectLimits/WarsongObjectInfo.cs
+++ b/src/WarcraftLegacies.Shared/FactionObjectLimits/WarsongObjectInfo.cs
@@ -9,8 +9,8 @@ public static class WarsongObjectInfo
   public static IEnumerable<ObjectInfo> GetAllObjectLimits()
   {
     yield return new(UNIT_O00C_GREAT_HALL_WARSONG_T1, Unlimited, UnitCategory.TownHall);
-    yield return new(UNIT_O02R_STRONGHOLD_WARSONG_T2, Unlimited);
-    yield return new(UNIT_O02S_FORTRESS_WARSONG_T3, Unlimited);
+    yield return new(UNIT_O02R_STRONGHOLD_WARSONG_T2, Unlimited, UnitCategory.TownHall);
+    yield return new(UNIT_O02S_FORTRESS_WARSONG_T3, Unlimited, UnitCategory.TownHall);
     yield return new(UNIT_O020_ALTAR_OF_CONQUERORS_WARSONG_ALTAR, Unlimited, UnitCategory.Altar);
     yield return new(UNIT_O01S_WAR_CAMP_WARSONG_BARRACKS, Unlimited, UnitCategory.Barracks);
     yield return new(UNIT_O009_REFINERY_WARSONG_RESEARCH, Unlimited, UnitCategory.Research);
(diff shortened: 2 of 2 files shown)


YakaryBovine:
@Technopigg Could I trouble you to review this as well? I'm keen for you to look at:
1. Whether the categories I assigned make sense
2. Whether this seems likely to break any of the unit replacement work you did (I don't think it should - I tested starting the game as both Warsong and Gilneas)

src/MacroTools.Shared/ObjectInfo.cs
```diff
@@ -10,18 +10,26 @@ public sealed class ObjectInfo
   public int Limit { get; }
 
   /// <summary>
-  /// The arbitrary category the object is assigned to, if any.
+  /// The arbitrary categories the object is assigned to, if any.
   /// </summary>
-  public UnitCategory Category { get; }
+  public List<UnitCategory> Categories { get; }
```
alrun3: Outside the scope of this PR, but just as a note; we should rework these to be general _identifiers_ that can be attached to objects for filtering, i.e. labels.

alrun3: Outside the scope of this PR, but just as a note; we should rework these to be general _identifiers_ that can be attached to objects for filtering, i.e. labels.

src/MacroTools.Shared/ObjectInfo.cs
```diff
@@ -10,18 +10,26 @@ public sealed class ObjectInfo
...
+    Categories = categorieses;
+    ObjectTypeId = objectTypeId;
+    Limit = limit;
+    LimitIncreaseHint = limitIncreaseHint;
+  }
+
   public ObjectInfo(int objectTypeId, int limit, UnitCategory category = UnitCategory.None, string? limitIncreaseHint = null)
   {
-    Category = category;
+    Categories = new List<UnitCategory> { category };
```
alrun3: Producing a list with a single UnitCategory of None is incompatible with the FactionSystem's UnitExtensions:

https://github.com/AzerothWarsLR/WarcraftLegacies/blob/8e4b017d5a70462edc2a181174b677f6a446bbe5/src/MacroTools/FactionSystem/UnitExtensions.cs#L31-L39

src/WarcraftLegacies.CLI/Migrations/UnitTooltipExtendedMigration.cs
```diff
@@ -56,7 +58,18 @@ private void DetermineTooltip(Unit unit)
   {
     var tooltipBuilder = new StringBuilder();
 
-    tooltipBuilder.Append(unit.TextTooltipExtended);
+    var unitId = (unit.NewId != 0 ? unit.NewId : unit.OldId).InvertEndianness();
+    var hasObjectInfo = _objectInfoRepository.TryGetObjectInfo(unitId, out var objectInfo);
+
+    if (hasObjectInfo && objectInfo.Categories.Count != 0)
+    {
+      AppendRoles(tooltipBuilder, unit, objectInfo);
```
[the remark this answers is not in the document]
YakaryBovine: It does yup. That is a bug with the implementation since I did intend for it to prepend the World Editor tooltip if there was no category, but on reflection I think I'll just configure every unit's category instead, and not ever use the World Editor tooltip. Sorry, probably put this PR up a little prematurely, got carried away!

src/MacroTools.Shared/ObjectInfo.cs
YakaryBovine: Oops - yeah that will cause issues if we ever do a replace and there is a preplaced unit with no configuration. Will fix.

Technopigg (commented):
Looking good , a nice clean finish for tooltips.

Categories themselves make good sense, clear and obvious what the role is; Destroyer threw me a little bit though. The odd unit may be in the wrong category (Kodo Beast).

In regards to the UnitPlacement work; there is an outstanding issue where if a preplaced units UnitCategory does not exist in the replacing factions ObjectInfo the unit will not be replaced, breaking the Faction script.  This is currently the case for Tauren (Tank) when Warsong is selected.

PR #3562, opened 2026-01-02
Add unit categories to units for all remaining factions

YakaryBovine:
Follow-up to #3561; adds unit categories to all remaining faction units...
Except boats. We'll deal with those later, I don't know what I want to do with them.

### DIFF
--- a/src/MacroTools.Shared/UnitCategory.cs
+++ b/src/MacroTools.Shared/UnitCategory.cs
@@ -10,7 +10,7 @@ public enum UnitCategory
   TownHall,
   GilneasManor,
   Shipyard,
-  LightHouse,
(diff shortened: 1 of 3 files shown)


alrun3 (commented):
Shouldn't there now be an unused overload that allows instantiation of objects without a category, which can be removed?

YakaryBovine:
> Shouldn't there now be an unused overload that allows instantiation of objects without a category, which can be removed?

I thought so too, but it turns out that the `ObjectInfo` object is used for all objects, not just units - and non-units don't use categories, so they use the zero-category version. It's probably not a great architecture.

alrun3 (approved):

PR #3563, opened 2026-01-02
Knowledge added to Ironforge quest rewards and quest discover in a linear order.

Technopigg:
Added Knowledge rewards to Ironforge quests, scaled by complexity.

- Quests now remain hidden until the previous quest or specific condition (training Magni, destroying Black Temple).
- Added Knowledge costs for relevant caster and specialist upgrades.
- Starting quest now unlocks multiple followup quests for a more linear progression.


mapdata/WarcraftLegacies/UpgradeData/R00V.json
```diff
@@ -33,7 +33,7 @@
     },
     {
       "Id": 1651338343,
-      "Value": 0
+      "Value": 5
```
alrun3: Don't these denote cost rather than reward?

PR #3569, opened 2026-01-03
Replace Rommath's barely functional Chain Mana Burn with our new custom one

YakaryBovine:
Rommath's current ult is called "Chain Mana Burn", but it's actually Ray of Neutralization with a mana drain component. It's a terrible ability with illogical targeting parameters; it only bounces to units with buffs, but is intended mainly for mana drain!

This replaces the ability with an actual Chain Mana Burn - the same one as Illidan has, but stronger.


alrun3 (approved):
Nice!

PR #3574, opened 2026-01-05
perf: use numeric constants in more places

alrun3:


alrun3:
Ready for review

YakaryBovine (approved):
Incredible, thank you for doing this, can't have been fun but it's so good to have it done.
I actually think the old Constants generator might not have had Items?

PR #3604, opened 2026-01-09
Illidan abandons the Broken Isles when he first travels to Outland

### LINKED ISSUE #3597: Remove Illidan's Broken Isles base when he first leaves
When Illidan first completes the Return to Outland quest, all of his units should be forcibly moved to Outland, with any buildings built refunded. This ensures he can't be reliant on his old stuff in the Broken Isles for balancing.

YakaryBovine:
Causes Illidan to abandon the Broken Isles completely when completing Return to Outland, refunding all buildings and teleporting all units to Outland. Actually affects all units on the map, just in case they make boats or something!

Closes #3597


Technopigg (approved):
All look good! Tried various ways to manipulate it and leave workers / units behind on Broken Isles with no success. Excellent work

PR #3608, opened 2026-01-10
Refactor unique unit naming system with NamePool & UnitNameManager 

Technopigg:
**Unique Elite Names 2.0**

- `NamePool` handles name assignment, release, and tracking of all names.
- `UnitNameManager` maps unit types to pools and delegates allocation logic.
- `UniqueNames` stores all unit type > name lists separately.
- Removes `EliteNameSystem`.
- Makes the system easier to extend for multiple name configurations in the future.
- Uses "ksv" over FourCC for flexibility.

### DIFF
--- a/src/WarcraftLegacies.Source/Setup/UniqueEliteNames.cs
+++ b/src/WarcraftLegacies.Source/Setup/UniqueEliteNames.cs
@@ -35,7 +35,7 @@ public static class UniqueEliteNames
           "Zann the Defender"
         },
       [UNIT_N00A_FARSTRIDER_QUELTHALAS_ELITE] =
-        new ()
+        new()
         {
           "Amora Eagleye",
           "Anthis Sunbow",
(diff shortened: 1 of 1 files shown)


alrun3 (commented):
Left some initial comments before looking at the implementation

src/WarcraftLegacies.Source/Setup/UniqueEliteNames.cs
```diff
@@ -0,0 +1,419 @@
...
+namespace WarcraftLegacies.Source.Setup;
+
+public static class UniqueEliteNames
+{
+  public static Dictionary<int, List<string>> GetNames()
+  {
+    return new Dictionary<int, List<string>>
+    {
+      [UNIT_H00F_PALADIN_LORDAERON] =
+        new List<string>
```
YakaryBovine: This one is mostly just for fun, but you don't need to specify the type here since it's already known. Same for the others below. Optional suggestion.

```suggestion
        new()
```

YakaryBovine (approved):
Nice job.

PR #3617, opened 2026-01-13
Reworked Deathsworn as powerful assassin units

### LINKED ISSUE #3615: Rework Illidari Deathsworn
Deathsworn are Illidari units trained from the Draenei Hut which currently don't have a function. Rework them to be Assassin class units that can be used alongside Akama to execute heroes and elites. Since they're plain units, they should feel all-in; good at reaching heroes and killing them, but once they do, are without a reliable way to avoid getting killed themselves. To that end, add the following abilities:

**Blink Strike:** Teleport to the target enemy unit and deal a small amount of damage (followed up by automatically attacking the target). High cooldown.
**Kingslayer:** Deal 40% additional damage to heroes and Resistant units.

Assassin units in general should feel like both a techtree and micro investment, so gate Kingslayer behind a Knowledge research and (perhaps) limit Deathsworn to 12.

YakaryBovine:
Deathsworn reworked
- Now have Blink Strike and Kingslayer
- Kingslayer costs 5 Knowledge
- Damage base 25 --> 40
- Damage sides per die 5 --> 7
- Attack delay 1.8 --> 1.4
- Limit unlimited --> 12
- Hit points 1230 --> 850

Closes #3615


Technopigg (commented):
All looks good, very cool units. 

My concern from a balance POV is that its an instant hero kill , if not more , in the early game. My suggestion would be to reduce the amount available to 6 at first , with the ability to later increase this to 12 to ensure they remain viable . Open to other suggestions also and just trying them out as is and balancing around.

PR #3646, opened 2026-01-28
Illidari Knowledge Upgrades & Narrative Quests

Technopigg:
**Quests**
- Added QuestAzshara
- Added QuestVestigesOfPower
- Added ThirstOfTheBetrayer

**Unit Changes** 

- Added Knowledge upgrades for Couatl and Scylla.
- Couatl and Scylla roles defined as Flying/AntiMage
- Siphon Mana added to Scylla , Feedback to Couatl's. 

**Buffed Parasite**

- Increased damage over time 4>15. 
- Spawned Parasites gain Lesser Parasite ability

### DIFF
--- a/mapdata/WarcraftLegacies/Imports/ReplaceableTextures/CommandButtonsDisabled/BTNFeedbackAlternate.blp
+++ b/mapdata/WarcraftLegacies/Imports/ReplaceableTextures/CommandButtonsDisabled/BTNFeedbackAlternate.blp
@@ -1,3 +1,0 @@
-version https://git-lfs.github.com/spec/v1
-oid sha256:337d593d9a678afce34210116b4e26b5bc8deea44170d65d96d1c5e97a132bad
-size 5197
--- a/mapdata/WarcraftLegacies/Imports/ReplaceableTextures/CommandButtonsDisabled/DISBTNFeedbackAlternate.blp
+++ b/mapdata/WarcraftLegacies/Imports/ReplaceableTextures/CommandButtonsDisabled/DISBTNFeedbackAlternate.blp
@@ -1,3 +1,0 @@
-version https://git-lfs.github.com/spec/v1
-oid sha256:06b60d07623eab1794b1cd7e42da240289c92ca5cd461dffaa7d8e519ac7e2e0
-size 3508


src/WarcraftLegacies.Source/Quests/Naga/QuestVestigesOfPower.cs
```diff
@@ -0,0 +1,67 @@
...
+      Artifacts.BookOfMedivh,
+      Artifacts.ScepterOfTheQueen,
+      Artifacts.AzureFragment,
+      Artifacts.EmeraldFragment,
+      Artifacts.RubyFragment,
+      Artifacts.ObsidianFragment,
+      Artifacts.BronzeFragment
+    };
+
+    var randIndex = (int)GetRandomReal(0, possibleArtifacts.Count - 0.01f);
```
YakaryBovine: No reason you would need a fractional number here.
```suggestion
    var randIndex = GetRandomInt(0, possibleArtifacts.Count);
```
Also, since `Count` goes up to 7 but the array only has items up to index 6, if this rolls 7, you will get an exception. You probably want: 
```suggestion
    var randIndex = GetRandomInt(0, possibleArtifacts.Count - 1);
```

Also also, you might consider weighting this, so that Book and Scepter are rolled more commonly than a Fragment. I just think that would be cooler, since the Fragments are the least cool thing here. Up to you though.

mapdata/WarcraftLegacies/AbilityData/Vkpt.json
```diff
@@ -0,0 +1,45 @@
...
+    {
+      "Id": 1718840929,
+      "Type": 3,
+      "Value": " (Scylla)"
+    },
+    {
+      "Level": 1,
+      "Id": 828536161,
+      "Type": 3,
+      "Value": "Absorbs the magical energies of a target enemy unit by taking \u003CVkpt,DataB1\u003E mana per second from it and giving it to this unit. |nLasts \u003CACsm,Dur1\u003E seconds."
```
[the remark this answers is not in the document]
YakaryBovine: Just checked, it can't, so probably need to update both the tooltip and the spell. Definitely not worth having Siphon Mana if it can't transfer mana to allies.

YakaryBovine (changes requested):
Nice job, just a few bits and bobs to update.

Technopigg:
Got stuck trying to circumvent `Launcher.IntegrityChecker.ObjectDataAccessibilityTests.AllUnits_CanBeTrained()` if you wouldn't mind letting me know the best method or what we do in that situation would be great !

mapdata/WarcraftLegacies/PathingMap.json
YakaryBovine: Looks like you don't have your TerrainArt file set up; this is causing the mountains to become pathable.

mapdata/WarcraftLegacies/Imports/ReplaceableTextures/CommandButtonsDisabled/BTNFeedbackAlternate.blp
```diff
@@ -0,0 +1,3 @@
+version https://git-lfs.github.com/spec/v1
```
YakaryBovine: Let's go ahead and delete these two files now that they're unused.

Technopigg:
requested changes made, excluding TerrainArt (having some issues)

YakaryBovine:
Superceded by https://github.com/AzerothWarsLR/WarcraftLegacies/pull/3648

PR #3690, opened 2026-02-07
Giving Rexxar Summon Hawk

IDeathlordI:
Made a new unit because of the Wanderer units that use the same ability, gave a different lore accurate name to it and fixed/updated tooltips based on the changes.

### DIFF
--- a/mapdata/WarcraftLegacies/AbilityData/A00G.json
+++ b/mapdata/WarcraftLegacies/AbilityData/A00G.json
@@ -55,7 +55,7 @@
       "Level": 1,
       "Id": 828536161,
       "Type": 3,
-      "Value": "Summons Spirit, a \u003Cn01J,realHP\u003E hit point hawk. Has the True Sight ability. |nLasts \u003Cn01J,Dur1\u003E seconds."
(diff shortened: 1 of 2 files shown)


mapdata/WarcraftLegacies/AbilityData/A00G.json
```diff
@@ -0,0 +1,70 @@
...
+      "Level": 1,
+      "Id": 829453409,
+      "Type": 3,
+      "Value": "Summon Spirit"
+    },
+    {
+      "Level": 1,
+      "Id": 828536161,
+      "Type": 3,
+      "Value": "Summons Spirit, a \u003Cnwe3,realHP\u003E hit point hawk. Has the True Sight ability. |nLasts \u003CANsw,Dur1\u003E seconds."
```
YakaryBovine: The duration here doesn't look right. It's still pulling from the now-deleted `ANsw` ability, whereas you want to be pulling from `A00G` for an accurate number.

I suspect using `nwe3` for hit points is wrong too; you want `n01J` for the new Spirit unit.

YakaryBovine (changes requested):
Nice fix, just some tooltip issues to resolve before I can merge.

mapdata/WarcraftLegacies/AbilityData/A00G.json
IDeathlordI: Checked in game and it looks like it's finally right

PR #3692, opened 2026-02-07
Several bug fixes

IDeathlordI:
Assumption is that all fountains of healing are now hostile and start with 2%(80)hp unless I missed some, the barracks near Stromgarde is now neutral passive. Ultravision no longer has a requirement for it to be researched and Summon Misha now only summons the one and only Misha.


mapdata/WarcraftLegacies/PathingMap.json
YakaryBovine: This is a funny one. All of the rocks in Legacies are unpathable due to custom edits we've made to our installations before saving the map. Since we forgot to tell you how to make those changes, all of the rocks in the map have become pathable, resulting in this 400mb file change!

I've put instructions for making your rocks unpathable in #essential-links on Discord: https://discord.com/channels/1161755984759378100/1161762028944379934

Once you've done that, you'll need to either:
1. Open the map again, resave it, and re-export. You should see your pathing changes here reverted
2. OR you can figure out how to drop this file from the commit entirely... Not sure how to do this in Github Desktop, ping me if you want to try figure it out

YakaryBovine:
> As this is now a business rule hardcoded in data, please add an integrity check to ensure that all current and future fountains are configured with the same percentage. Alternatively, use the PreplacedUnits scanner to iterate over all fountains and adjust them at runtime

Sounds like a good idea, but this fix was only flagged as requiring World Editor knowledge and there are probably easier issues for Deathlord to get his feet wet with, so let's call this an optional requirement for the PR.

YakaryBovine (approved):
Looking good. There are easier ways to drop a file from a commit, but your revert-then-re-add technique is just fine for now.

Yep, the Fountain of Blood is a separate issue so it's good to leave it alone here.

PR #3693, opened 2026-02-08
Add 3 new Knowledge researches to Illidari's Spawning Grounds

YakaryBovine:
Keen to get a basic balance review on this one. Do the researches look useful, do they all look worth 5 Knowledge? Doesn't have to be super precise, just a vibe.

image
image
image


YakaryBovine:
Added building requirements to Improved Slow Poison and Heavy Boulders that correspond with the unit requirements.

PR #3698, opened 2026-02-09
Make Illidan's portal from the Black Temple one-way

### LINKED ISSUE #3686: Fel Horde can get units into the Broken Isles prior to Illidan completing "Return to Outland"
1. As Illidan, open a portal to Black Temple from the Broken Isles at the start of the game
2. As Fel Horde, send workers through this portal
3. As Illidan, walk through to Outland to complete "Return to Outland"
4. Illidan's units are forcibly teleported to Outland, but Fel Horde's are left behind, leaving them free to creep the Isles

Only an issue in `dev`.

YakaryBovine:
Closes #3686

Closes up some gaps by changing how Illidans' initial portal to Outland works. Now, when completing the Broken Isles, he gets a static one-way portal there. 
image

When he goes through, the quest completes, unlocking his proper Portal to Black Temple ability and moving all of his units from the Broken Isles as before.


YakaryBovine:
> Very nice , all works correctly from what I tested. Might it be worth spawning the portal near Illidans current location as you'll never be looking at the middle at the time the quest completes. Just a thought

Yeah sure, I'll have a play around. Just thought putting in front of that arch looked cool. Alternatively I might force the camera to the portal if the portal spawning on Illidan doesn't look cool.

Issue #3688, opened 2026-02-07
War Stomp can stun air units


YakaryBovine:
Yep, definitely a bug.

PR #3724, opened 2026-02-14
Custom Stomp.cs spell balancing/fixing

Chuckaay:
The purpose of this pull request is to fix and balance the custom stomp spell. This includes the following changes:

1. Create 2 new cast filters in CastFilters.cs; one cast filter is designed to only target enemy ground units (doesn't include structures or flying units), and the other is designed to target enemy units (doesn't include structures, but does include flying units)
2. Changes to cast filtering in Stomp.cs so that by default it only targets enemy ground units. Furthermore allow for it to be configured in the case where a spell using this will be allowed to target more unit types (e.g. Flying).
3. Adjust some existing Stomp.cs spells so that they are able to target flying units, based on balance feedback.

Screenshots of the changes in action:

Cairne's War Stomp now behaves like vanilla stomp and uses the default setting of Stomp.cs (only targets enemy ground units)

_Damaging and stunning ground units:_
image

_Doesn't damage or stun flying units or structures:_
image

Mammoth Wrangler's Crippling Shout also uses Stomp.cs but based on balance feedback should be allowed to also hit flying units

_Debuffing ground units:_
image

_Debuffing flying units, and not structures_
image

This will resolve issue #3688

### DIFF
--- a/src/MacroTools/Spells/CastFilters.cs
+++ b/src/MacroTools/Spells/CastFilters.cs
@@ -35,13 +35,6 @@ public static class CastFilters
   /// <summary>
   /// Used for spells where you only want them to affect enemy ground units. Flying units and structures are not affected.
   /// </summary>
-  public static bool IsTargetEnemyAliveAndGround(unit caster, unit target)
-  {
-    var casterPlayer = caster.Owner;
-    return caster != target &&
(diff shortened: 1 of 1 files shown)


YakaryBovine (dismissed):
Great work, and thanks for demonstrating the testing so thoroughly, that makes my job much easier.

Approved; we'll get this merged in during the first minor version after the Stormrage patch (since Stormrage is now feature complete).

PR #3758, opened 2026-02-15
Forfeit command qol

Ripzss:


YakaryBovine (changes requested):
Literally just a full stop to fix, then can merge.

PR #3773, opened 2026-02-16
Bug fixes

IDeathlordI:
Elites now can be trained from Drak'tharon Keep.

Tooltip for Shadow Aura has been fixed.

Ghouls can no longer attack air.

Battle Station no longer requires Pillaging Echo Isles.

Fel Warlock Adept Training can no longer be researched from Kil'sorrow Fortress.

Conjurer's invisibility's tooltip has been fixed.

Vortex Pinnacle(Skywall's capital)'s collision size has been increased, which means melee units can now attack it.


YakaryBovine (approved):
Perfect

PR #3777, opened 2026-02-17
Normalization of Casters' initial hit points

IDeathlordI:
700hp for unlimited casters for 3p factions.

900hp for limited casters for 3p factions.

800hp for unlimited casters who have 3 research upgrades instead of the normal 2(Dala casters,Dwarf engineer etc) for 3p factions.

600hp for Sunfury's unlimited casters since they have the mana shield mechanic.

800hp for Sunfury's limited casters since they have the mana shield mechanic.

1000hp for The Exodar's unlimited casters since they are unique and pretty expensive.

These general changes have been done on all normal factions, alternative factions(Warsong,etc) and paths(Scarlet Crusade).

All the tooltips related to the casters' researches have been updated.

### DIFF
--- a/mapdata/WarcraftLegacies/UpgradeData/R000.json
+++ b/mapdata/WarcraftLegacies/UpgradeData/R000.json
@@ -113,13 +113,13 @@
       "Level": 1,
       "Id": 828536167,
       "Type": 3,
-      "Value": "Increases Tidesages\u0027 mana capacity, mana regeneration rate and gives them the ability to cast Summon Sea Elemental."
+      "Value": "Increases Tidesages\u0027 mana capacity and mana regeneration rate, and gives them the ability to cast Summon Sea Elemental."
     },
     {
       "Level": 2,
       "Id": 828536167,
       "Type": 3,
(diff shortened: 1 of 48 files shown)


mapdata/WarcraftLegacies/UpgradeData/R000.json
```diff
@@ -108,13 +113,13 @@
       "Level": 1,
       "Id": 828536167,
       "Type": 3,
-      "Value": "Increases Tidesages\u0027 mana capacity, mana regeneration rate, hit points and gives them the ability to cast Summon Sea Elemental."
+      "Value": "Increases Tidesages\u0027 mana capacity, mana regeneration rate and gives them the ability to cast Summon Sea Elemental."
```
YakaryBovine: Just a grammar fix here...
```suggestion
      "Value": "Increases Tidesages\u0027 mana capacity and mana regeneration rate, and gives them the ability to cast Summon Sea Elemental."
```
I think you should be able to use find and replace to make a similar fix across all of the other upgrades, you shouldn't need to do it manually per.

Issue #3607, opened 2026-01-10
Convention-based GameSetup via Reflection


alrun3:
`GameSetup` is currently ad-hoc, requiring explicit callsites and boilerplate to wire initialization logic.

`CSharp.lua` provides basic reflection support to scan assemblies and invoke methods by name. This could be leveraged to introduce convention-based initialization; for example, automatically invoking all methods matching a known signature such as:

```csharp
public static void OnGameSetup()
```

This approach is intentionally minimal, aiming to reduce ceremony around initializers and provide a single callsite that can support cross-cutting concerns (e.g. profiling).

If this proves insufficient for managing interdependencies, a follow-up convention (e.g. a post-initialization hook) could be added

YakaryBovine:
@alrun3 I completely agree; vertically slicing by faction would be a great folder architecture for this map because the vast majority of end-user features are oriented around factions. Seems like a great change.

The only caveat is that the `FactionObjectLimits` folder is in the Shared project because it needs to be exposed to `WarcraftLegacies.CLI`; it's used in migrations to help determine unit tooltips.

PR #3871, opened 2026-03-15
Reorganize WarcraftLegacies.Source by faction-based vertical slices

YakaryBovine:
A step towards #3607. Moves content that is specific to individual factions into folders named after those factions. There should be no functional changes or refactors to individual files, other than renames of classes.


YakaryBovine:
> LGTM, nice to have normalized all the namespaces too, e.g. `Fel_Horde` vs. `FelHorde`. Maybe we'll trim some more in the future - e.g. is the `Factions` namespaces needed?

Yeah, that would be nice. I've gotten so used to enforcing folders = namespace but there are lots of cases in this project where having a shorter namespace than the folder structure would be convenient. I wonder if there's a way to truncate specifically just the `Factions` part of the namespace while leaving the rule in tact?

> What about spells? Since we've made some changes in this PR, should we move the remaining ones to `Shared`, and those that only have a single reference, and aren't generic, move them to a faction-specific `Spells` folder (e.g. `ExactJusticeSpell`)?

Yeah, definitely. I think I'll do that in a follow-up PR.

PR #3878, opened 2026-03-19
Vampirika/hardened plating standardisation

vampirika:
TODO: 
- Research doesn't seem to be upgrading from level 1 to level 2 e.g. legion (base) upgrade in barracks should increase from 6 to 12 on fel guard.
- Create/use hardened research icon (use Mountain Giant one).
- Add researches to all other factions that require.
- Check if Hardened skill position fits on all units.


YakaryBovine:
@vampirika Before you get too far: this is definitely a `dev`-targeted change in complexity!

PR #3885, opened 2026-03-21
Feat/starting resources

### LINKED ISSUE #3882: Give player starting gold at 0 seconds rather than 60 seconds
This is so that players can queue up their starting buildings during the starting cinematic rather than having to rush to do it at the start of turn 1.

alrun3:
* Starting resources are now given when a player is assigned to a faction
* Starting resources for Scarlet Crusade are now given in the same way as other factions
* Factions are now only added to the multiboard using events

Closes #3882


YakaryBovine (approved):
Nice, that makes more sense. I especially like that optional factions properly get their starting gold and income even when I skip the intro cinematic before they're picked.

Issue #3886, opened 2026-03-22
Silvermoon Control Point could not be recaptured after being captured.


Technopigg:
Dev  branch.

SIlvermoon CP when taken by Scourge was unable to be retaken , income also appealingly stopped being received from it.

PR #3931, opened 2026-03-29
fix: doomhammer tooltip references `YBRR` instead of `YBRO`

alrun3:
Split from #3928


YakaryBovine:
Dang, wish I had looked at this before doing a hotfix!

YakaryBovine:
> > Dang, wish I had looked at this before doing a hotfix!
> 
> My bad, I didn't want to tag you with notifications at night

Not your fault! Also it turns out I have to do another hotfix #3886 anyway, so this will go out with that.

PR #3949, opened 2026-03-31
Chore/unused buffs

alrun3:
* Integrity checker now validates references to buffs on fields we've modified
* All abilities now reference valid buffs
* All buffs reported as unused are removed
* Fixes some upstream issues with buffs, which also manifested in our files. E.g. BUhf (non-existent) vs. Buhf (exist). We can choose to revert these

TODO: 
* Follow up on A09N and the use of `ModAbilityAvailability`

If reviewed OK, can I ask for you to merge and follow up with the cleanup of models? I don't have Git LFS configured


src/Warcraft.Integrity/UnreachableObjectCollection.cs
```diff
@@ -1230,6 +1239,27 @@ private void RemoveWithChildren(Ability ability)
...
+        }
+
+        try
+        {
+          foreach (var buff in ability.StatsEffects.TryGetBuffsAtLevel(i, ability.IsStatsEffectsModified[i]))
+          {
+            RemoveWithChildren(buff);
+          }
+        }
+        catch (InvalidCastException) when (ability is DeathKnightAnimateDead)
```
YakaryBovine: Never seen this syntax before, containing a catch with a type check. Cool

PR #3959, opened 2026-04-01
chore: remove unused abilities only disabled in code

alrun3:


YakaryBovine (approved):
Wow, there were quite a few of these.

PR #3963, opened 2026-04-02
chore: remove extraneous ability modifications

alrun3:
Applies some of the learnings from https://github.com/AzerothWarsLR/WarcraftLegacies/pull/3960. We'd want to automate this in the future, but for now let's just prune the excess

* Removes object modifications that exceed the level of the object
* Removes object modifications that belong to hero abilities on non-hero abilities


YakaryBovine (approved):
Nice. It's pretty funny that the World Editor doesn't do this on its own!

PR #4019, opened 2026-04-16
Refund System

### LINKED ISSUE #4009: Set terrain in Base Spawn locations to be "unbuildable" prior to the event firing
Players are currently able to pre‑build towers and other structures in locations where bases are later spawned by quests or where armies are teleported after completing a quest. These areas are intended to be reserved for scripted events, but the game does not currently prevent players from placing buildings there.

As a result, bases may spawn on top of player‑built structures, or teleport events may occur inside obstructed areas. This behaviour is unintended.

Technopigg:
A small refund system as a alternative solution to close #4009.

- Enemy buildings located in a base spawn region are removed on the quests completion
- Players are refunded resources for all removed buildings
- Players are informed of the removal and refund, including the total refunded amount.


Technopigg:
Refund system as discussed in discord 

@alrun3 can you review the code
@vampirika if the codes all good do you mind testing in game?

alrun3:
@Technopigg any chance we could not target `main` for this? It'll mean another round of merge conflicts

Technopigg:
Yup no prob at all. Apologies I targeted main thinking there *wouldn't* be any conflicts 🙄

PR #4021, opened 2026-04-17
Refund System + Unused Model Removal

### LINKED ISSUE #4009: Set terrain in Base Spawn locations to be "unbuildable" prior to the event firing

Technopigg:
A small refund system as a alternative solution to close #4009 

Enemy buildings located in a base spawn region are removed on the quests completion
Players are refunded resources for all removed buildings
Players are informed of the removal and refund, including the total refunded amount.


src/WarcraftLegacies.Source/Regions.g.cs
alrun3, line 1: "Aerie Peak Copy" was deleted so constants need to be regenerated. We should probably add an automated test to check for diffs in these files

src/WarcraftLegacies.Source/Factions/Stormwind/Quests/QuestStromgarde.cs
```diff
@@ -52,13 +52,23 @@ protected override void OnFail(Faction completingFaction)
...
+
     completingFaction.Player.SetTechResearched(ResearchId, 1);
+
     foreach (var unit in _rescueUnits)
     {
       unit.Rescue(completingFaction.Player);
     }
   }
 
+
```
alrun3: Are these double spaces between method declarations deliberate?

src/WarcraftLegacies.Source/Factions/Legion/Quests/QuestArgusControl.cs
```diff
@@ -43,6 +43,8 @@ public QuestArgusControl() : base("Argus",
   /// <inheritdoc/>
   protected override void OnComplete(Faction completingFaction)
   {
+    DisplayTextToPlayer(GetLocalPlayer(), 0, 0, "Argus quest OnComplete fired");
+
```
alrun3: Looks like a leftover debug log

alrun3 (approved):
Overall functional when merged with `dev`, so feel free to merge after, however some feedback to consider:

The method name doesn't convey that it refunds structures not owned by the player or neutral players. I think we should make that clearer in the name, but saying that also makes me wonder whether we intended for allied structures to also be refunded? If not we should use `IsUnitEnemy` instead, which respects alliances rather than just ownership.

The method also makes callers responsible for enumerating units before passing them in, that detail could be handled internally. The multi-region case in `QuestPlague` results in three separate calls where one could do. Since the accumulation only exists to support this case, we can solve this at the signature level as well.

Proposed method signatures:

```csharp
public static void RefundEnemyStructuresInRect(player whichPlayer, params Rectangle[] regions)
public static void RefundEnemyStructuresInRange(player whichPlayer, float x, float y, float radius)
```

Happy to help you implement

src/WarcraftLegacies.Source/Factions/Legion/Quests/QuestArgusControl.cs
Technopigg: oops it is indeed

src/WarcraftLegacies.Source/Factions/Stormwind/Quests/QuestStromgarde.cs
Technopigg: They are not intentional, nope. just habit I think. I would of thought `dotnet format` would of cleaned that up but perhaps I am asking a bit much :D

src/WarcraftLegacies.Source/Regions.g.cs
Technopigg: gotcha and makes good sense yup!

src/WarcraftLegacies.Source/Factions/Stormwind/Quests/QuestStromgarde.cs
alrun3: I'll see if it's possible to add that rule. Feel free to remove the excess newlines from the PR :+1:

Technopigg:
> Overall functional when merged with `dev`, so feel free to merge after, however some feedback to consider:
> 
> The method name doesn't convey that it refunds structures not owned by the player or neutral players. I think we should make that clearer in the name, but saying that also makes me wonder whether we intended for allied structures to also be refunded? If not we should use `IsUnitEnemy` instead, which respects alliances rather than just ownership.
> 
> The method also makes callers responsible for enumerating units before passing them in, that detail could be handled internally. The multi-region case in `QuestPlague` results in three separate calls where one could do. Since the accumulation only exists to support this case, we can solve this at the signature level as well.
> 
> Proposed method signatures:
> 
> ```cs
> public static void RefundEnemyStructuresInRect(player whichPlayer, params Rectangle[] regions)
> public static void RefundEnemyStructuresInRange(player whichPlayer, float x, float y, float radius)
> ```
> 
> Happy to help you implement

Sounds like a solid plan. I refactored the original method I had into a `MacroTools` class any didn't quite nail the naming. The previous class was called RefundEnemyStructures. I guess I overlooked it with this one! and cool all makes solid  sense as far I can tell. I think I should be okay making those changes but will give you a shout if I get stuck! tyvm

Technopigg:
Appears that I merged the unused models back in also 

  `Expected unusedModels to be empty because every imported model should be used by an ability, doodad, buff, destructable, or script. Found 3, but found {"markofmagic.mdx", "war3mapimported\heaven's gate channel.mdx", "war3mapimported\sacredshield.mdx"}.`

alrun3:
> Appears that I merged the unused models back in also
> 
> `Expected unusedModels to be empty because every imported model should be used by an ability, doodad, buff, destructable, or script. Found 3, but found {"markofmagic.mdx", "war3mapimported\heaven's gate channel.mdx", "war3mapimported\sacredshield.mdx"}.`

You never removed them: https://github.com/AzerothWarsLR/WarcraftLegacies/pull/4018 :smile:

Technopigg:
> > Appears that I merged the unused models back in also
> > `Expected unusedModels to be empty because every imported model should be used by an ability, doodad, buff, destructable, or script. Found 3, but found {"markofmagic.mdx", "war3mapimported\heaven's gate channel.mdx", "war3mapimported\sacredshield.mdx"}.`
> 
> You never removed them: #4018 😄

XD haha what! how did I manage to Close instead of Merge! ty good catch

src/MacroTools/Utils/RefundSystem.cs
```diff
@@ -0,0 +1,64 @@
...
+
+namespace MacroTools.Utils;
+
+public static class RefundSystem
+{
+  private static readonly Dictionary<player, float> _globalTotals = new();
+
+  private static void RefundHostileStructures(player whichPlayer, IEnumerable<unit> units)
+  {
+    foreach (var u in units.Where(u => u.IsUnitType(unittype.Structure) && u.IsEnemyTo(whichPlayer)))
```
alrun3: Did we also have to filter neutral player structures?

Technopigg: Neutral Passive doesnt count as enemy , but do you think we should for safety?

alrun3: Only if it serves a purpose, I just recall you had included them in the filter before

alrun3 (approved):

src/MacroTools/Utils/RefundSystem.cs
Technopigg: gotcha , I think I was just being over cautious before. There will never be a situation where neutral hostile structures will be in one of these locations.

vampirika: Control points are also removed, so need to exclude them

src/WarcraftLegacies.Source/Factions/FelHorde/Quests/QuestRebuildBlackTemple.cs
```diff
@@ -32,6 +33,12 @@ public QuestRuinsofShadowmoon(Rectangle rescueRect) : base("Ash and Smoke",
   /// <inheritdoc/>
   protected override void OnComplete(Faction whichFaction)
   {
+    var p = whichFaction.Player;
```
vampirika: This should be in the quest "Blackrock" not black temple, and probably not needed in this quest

src/MacroTools/Utils/RefundSystem.cs
alrun3: Ah yes, I am generally not able to review anything related to content as I don't play the game. You should check `IsRemovable`, a variant of it or check for `unittype.Ancient` (IIRC): https://github.com/AzerothWarsLR/WarcraftLegacies/blob/286067b1a3479581d478cfeefaa25f8392dab90b/src/MacroTools/Extensions/UnitExtensions.cs#L444-L453

alrun3:
Consider using squash-merge for this PR. Several commits either fail tests or lack meaningful context in their messages, e.g.:

* `RefundEnemyStructures`
* `Update RefundSystem.cs`
* `Delete RefundEnemyStructures.cs`
* `Merge branch 'dev' into Technopig-PlayerRefund`

vampirika (changes requested):

src/WarcraftLegacies.Source/Factions/FelHorde/Quests/QuestRebuildBlackTemple.cs
Technopigg: oops good catch

vampirika (approved):
Looks good

PR #4031, opened 2026-04-18
Fixing hardened values and tooltips

vampirika:
updated tooltip for unholy plating (death knight)
reduced illidari elite from 18 to 12 (no upgrade)
increased war plating for felguard 6 to 12 (requires upgrade)
added new ability for doomguard to use limited values (15)


Technopigg (approved):
All solid

PR #4044, opened 2026-04-21
Technopig Ironforge 4.8.0, Researches, Units and unit Abilities 

### LINKED ISSUE #4038: Ironforge Upgrades
We need to rework any existing upgrades that Ironforge has that may not be worth the Knowledge investment and those that are unlikely to ever be researched. 

There should be a focus on the player not "needing" every research to perform well but being able to select a handful of researches from a larger pool that work effectively. 

The roster should feel like it improves with the upgrades meaning that they should be relevant to Ironforge's playstyle.

Technopigg:
- [x] Thane into a normal unit and related stats and abilities
- [x] Engineer into elite unit with stats and abilities 
- [x] Research for Dreadnaught
- [x] Research for Thane 
- [x] Rune Smith added
- [x] Rune Priest updated
- [x] Researches for Rune Priest
- [x] Research choices for Steam Tank, Flamethrower and Artillery
- [x] Drunken Haze ability for Rune Priest 
- [x] Flamethrower ability for Flame Tank 
- [x] Volatile Flask (AoE damage , applies Burning debuff to those under the effects of Drunken Haze )
- [x] Uniform unit scaling
- [ ] Balance check on all new units and abilities 

To Fix


vampirika:
As mentioned on discord, I seem to be missing some icons and the model for runepriest, but that might be on me. So will leave that for now. 

**From what I could see:**
Some hotkeys need updating (might be done through migration?)

Engineer
- Engineer statis trap C instead of E
- Engineer turret states it has 750hp but only has 200hp

Thane
- Training Thane was W instead of R
- Thane could be made stronger hp (similar to mountain giants).
- Thane W upgraded didn't seem to work? Called "rounds for everyone"

Magic
- Runesmith Q now has effect that looks like aoe, I don't think it should have this as a 'bolt'. Could be put onto E instead?
- The flame breath from the E didn't always seem to apply the DoT to drunken haze target?

Siege
- Does dreadnought upgrade need an icon? We don't normally do it for simple stat increases like hp and ms.
- Tank still has attack delay in tank/artillery form. Maybe a bit weird?
- Also, tank has no true projectile. Might be okay, again just a bit unusual.

**Misc:**

War golem has mana but no mana-using ability. Can be removed.
Warrior probably doesn't need forgeworks requirement anymore?


Mostly minor things it looks like. Seem really good otherwise 😃

Technopigg:
> As mentioned on discord, I seem to be missing some icons and the model for runepriest, but that might be on me. So will leave that for now.
> 
> **From what I could see:** Some hotkeys need updating (might be done through migration?)
> 
> Engineer
> 
> * Engineer statis trap C instead of E
> * Engineer turret states it has 750hp but only has 200hp
> 
> Thane
> 
> * Training Thane was W instead of R
> * Thane could be made stronger hp (similar to mountain giants).
> * Thane W upgraded didn't seem to work? Called "rounds for everyone"
> 
> Magic
> 
> * Runesmith Q now has effect that looks like aoe, I don't think it should have this as a 'bolt'. Could be put onto E instead?
> * The flame breath from the E didn't always seem to apply the DoT to drunken haze target?
> 
> Siege
> 
> * Does dreadnought upgrade need an icon? We don't normally do it for simple stat increases like hp and ms.
> * Tank still has attack delay in tank/artillery form. Maybe a bit weird?
> * Also, tank has no true projectile. Might be okay, again just a bit unusual.
> 
> **Misc:**
> 
> War golem has mana but no mana-using ability. Can be removed. Warrior probably doesn't need forgeworks requirement anymore?
> 
> Mostly minor things it looks like. Seem really good otherwise 😃

Nice catches! No idea how Burning Oil found its way onto Rune of Wrath :S. Swapped out or Firebolt and fixed the rest. 

Can buff Thanes if you like , I loosely based the stats off of Mountain Giant too. My concern with them is that they are not really needed it seems and will be just used for macro and be quite busted at it. But I suppose you could do that with MGs.

The assets maybe due to some LFS stuff I am attempting to sort at the moment, not 100% sure.

PR #4047, opened 2026-04-23
Technopig Scourge Flavour

Technopigg:
Added an Extension for RescueGroup that allows Crypt Fiends to "unburrow" when Rescued.


src/WarcraftLegacies.Source/Factions/Scourge/Quests/QuestPlague.cs
```diff
@@ -186,12 +187,28 @@ private void SpawnArmies(Faction completingFaction)
...
         {
-          if (!unit.IsUnitType(unittype.Peon))
+          if (parameter.SummonUnitTypeId == UNIT_UCRY_CRYPT_FIEND_SCOURGE)
           {
-            unit.IssueOrder(ORDER_ATTACK, attackTarget.X, attackTarget.Y);
+            u.SafelyRemove();
+            var burrowed = unit.Create(primaryPlaguePlayer, UNIT_UCRM_BURROWED_CRYPT_FIEND_SCOURGE,
+              position.X, position.Y, 0);
+
+            var timer = new Timer(_ =>
```
vampirika: I believe this is causing a loop and spiders at the plague are getting constant orders. Might want a break instead of continue?

mapdata/WarcraftLegacies/Units/2048_19968.json
```diff
@@ -1,4 +1,44 @@
 [
+  {
```
vampirika, line 2: I think this unit is Scourge instead of Hostile Passive. It's one in Drak Tharon Keep

vampirika:
Burrowed crypt fiends look smaller than when they are unburrowed. I think the scaling doesn't match. Other than that and the other minor comments, it's a nice touch

vampirika (changes requested):

Technopigg:
> Burrowed crypt fiends look smaller than when they are unburrowed. I think the scaling doesn't match. Other than that and the other minor comments, it's a nice touch

You mean the small splat texture they show when burrowed?

vampirika:
> > Burrowed crypt fiends look smaller than when they are unburrowed. I think the scaling doesn't match. Other than that and the other minor comments, it's a nice touch
> 
> You mean the small splat texture they show when burrowed?

Nah whole unit, if you see it will get bigger once it's finished animation

src/WarcraftLegacies.Source/Factions/Scourge/Quests/QuestPlague.cs
Technopigg: I don't think `continue` causes a loop, just skips the rest of the inner. afaik this timer should just fire the lambda once after 0.15 secs and shouldn't loop loop. I am not 100% though.

mapdata/WarcraftLegacies/Units/2048_19968.json
Technopigg: yh it was nice one

Technopigg:
> > > Burrowed crypt fiends look smaller than when they are unburrowed. I think the scaling doesn't match. Other than that and the other minor comments, it's a nice touch
> > 
> > 
> > You mean the small splat texture they show when burrowed?
> 
> Nah whole unit, if you see it will get bigger once it's finished animation

ah yeah scale was 0.2 lower than the Crypt Fiend

src/WarcraftLegacies.Source/Factions/Scourge/Quests/QuestPlague.cs
Technopigg: @alrun3 small one but what do you recommend regarding the above?

alrun3: Haven't looked at the WCSharp timer internals, but based on this call site it seems like these timers are interleaved (similar to [TimerQueue](https://www.hiveworkshop.com/threads/timerqueue-stopwatch.353718/)), which would cause the callback to fire repeatedly every 0.15s (issuing orders on every tick). To make this a one-shot timer, set `t.Active = false` in the callback:

```diff
diff --git a/src/WarcraftLegacies.Source/Factions/Scourge/Quests/QuestPlague.cs b/src/WarcraftLegacies.Source/Factions/Scourge/Quests/QuestPlague.cs
index 3d0d7ed87..cacd31d03 100644
--- a/src/WarcraftLegacies.Source/Factions/Scourge/Quests/QuestPlague.cs
+++ b/src/WarcraftLegacies.Source/Factions/Scourge/Quests/QuestPlague.cs
@@ -196,13 +196,12 @@ public sealed class QuestPlague : QuestData
             var burrowed = unit.Create(primaryPlaguePlayer, UNIT_UCRM_BURROWED_CRYPT_FIEND_SCOURGE,
               position.X, position.Y, 0);
 
-            var timer = new Timer(_ =>
+            TimerSystem.Add(new Timer(t =>
             {
               burrowed.IssueOrder(ORDER_UNBURROW);
               burrowed.IssueOrder(ORDER_ATTACK, attackTarget.X, attackTarget.Y);
-            }, 0.15f);
-
-            TimerSystem.Add(timer);
+              t.Active = false;
+            }, 0.15f));
             continue;
           }
```

vampirika (dismissed):
Just need test fix, otherwise good

src/WarcraftLegacies.Source/Factions/Scourge/Quests/QuestPlague.cs
Technopigg: I had a look at WCSharp.Timers and I _think_ its okay. The `Action` method hasnt got anything that requeues or schedules more ticks afaik.  I think it would need `Repeats = True` for that 
```
csharp
public void Action()
{
    if (!this.Active)
        return;
    this.TimerAction(this);
}```

and the constructor looks okay too. 

```
```
csharp
public Timer(Action<Timer> timerAction, float timeout)
{
    this.TimerAction = timerAction;
    this.Timeout = timeout;
    this.root = TimerSystem.GetOrCreate(timeout);
}
```

So am I right in assuming that `t.Active = false` here would be redundant as the timer already is non repeating and self terminating?

Let me know if any of that is wildly off though :D

Technopigg:
> Just need test fix, otherwise good

for some reason it thinks there is a model in this branch which actually isn't anymore.  :S

src/WarcraftLegacies.Source/Factions/Scourge/Quests/QuestPlague.cs
alrun3: Ah, apologies. I read Vampirika's comment as an observation and not a question, and inferred the behavior from that.

Yep, then the implementation would be fine! It would also match the style of other WCSharp systems, where when you add an instance to the system, the system sets `Active` to `true`

src/WarcraftLegacies.Source/Factions/Scourge/Quests/QuestPlague.cs
```diff
@@ -186,12 +187,28 @@ private void SpawnArmies(Faction completingFaction)
...
-        foreach (var unit in CreateUnits(primaryPlaguePlayer, parameter.SummonUnitTypeId,
+        foreach (var u in CreateUnits(primaryPlaguePlayer, parameter.SummonUnitTypeId,
                    position.X, position.Y, 0, parameter.SummonCount))
         {
-          if (!unit.IsUnitType(unittype.Peon))
+          if (parameter.SummonUnitTypeId == UNIT_UCRY_CRYPT_FIEND_SCOURGE)
           {
-            unit.IssueOrder(ORDER_ATTACK, attackTarget.X, attackTarget.Y);
+            u.SafelyRemove();
+            var burrowed = unit.Create(primaryPlaguePlayer, UNIT_UCRM_BURROWED_CRYPT_FIEND_SCOURGE,
```
alrun3: Why are we not creating `UNIT_UCRM_BURROWED_CRYPT_FIEND_SCOURGE` instead of `UNIT_UCRY_CRYPT_FIEND_SCOURGE`? We wouldn't have to remove and re-create the units then

alrun3:
> > Just need test fix, otherwise good
> 
> for some reason it thinks there is a model in this branch which actually isn't anymore. :S

It was in `dev` at the time of the test run. Can you try to re-run it?

src/WarcraftLegacies.Source/Factions/Scourge/Quests/QuestPlague.cs
Technopigg: ha good point :D

Technopigg:
> > > Just need test fix, otherwise good
> 
> > 
> 
> > for some reason it thinks there is a model in this branch which actually isn't anymore. :S
> 
> 
> 
> It was in `dev` at the time of the test run. Can you try to re-run it?

No change when re running the checks . Odd

src/WarcraftLegacies.Source/Factions/Scourge/Quests/QuestPlague.cs
vampirika: This is still bugged. Crypt fiends from plague are stuck in a loop

vampirika (dismissed):
Test is fixed which is great and scale is now fine, but crypt fiends from plague is still stuck in a loop

Technopigg:
> Test is fixed which is great and scale is now fine, but crypt fiends from plague is still stuck in a loop

Ah I see what you mean now. So it ended up being the Attack Order being issued before the Un-burrow animation had finished creating a broken order queue. Tested now all working correctly.

alrun3 (approved):
LGTM

PR #4053, opened 2026-05-26
Update github-releases-to-discord.yml

Technopigg:


.github/workflows/github-releases-to-discord.yml
```diff
@@ -15,6 +15,7 @@ jobs:
           color: "2105893"
           username: "Warcraft Legacies Releases"
           avatar_url: "https://cdn.discordapp.com/avatars/487431320314576937/bd64361e4ba6313d561d54e78c9e7171.png"
+          embed_title: "Download Link: ${{ github.event.release.tag_name }}"
```
alrun3, line 18: I couldn't find `embed_title` in the [documented configuration options](https://github.com/SethCohen/github-releases-to-discord/tree/v1.13.1#configuration) for `SethCohen/github-releases-to-discord@v1.13.1`. Are you sure it’s valid?

Technopigg: whoops , yh you are completely correct. I may consider changing the action then to this one https://github.com/Ilshidur/action-discord

Mainly just as people constantly say they cant find the link with the current format

Technopigg:
I think the better option here is just renaming the releases to "Download 3.8.0" for example

PR #4054, opened 2026-06-13
Ironforge polish and some balance

vampirika:
Added balance and polish to Ironforge, including:

Standardised Breath of fire burn to 20dps
Buffed thane hp
Removed stormrider gryphon unit
Checked engineer mana costs/values
Reduced Dagran costs and added more intelligence
Changed kalimdor expedition quest northern option from auberdine (player locked) to stonetalon peak region.
Reduced deeprun tram gold cost
Allowed thanes to be trained from shadowforge city and aerie peak (flavour of "thanes" being clan leaders)


Technopigg (approved):
As discussed on Discord all looks good. 
Will raise a ticket for the changed needed to Falstad's Stormguard ability as a result.

PR #4055, opened 2026-06-15
feat: add Muradin's Stormbolt spell

alrun3:
Split from #4051 

The new icons are omitted. We can add them when we merge Muradin's toolkit, which contains the paths and assets.

As per offline discussion:

If we don't like how the ability behaves, we can consider the Forked Lightning approach instead, but since I already had most of the code in place, I opted for trying the straight line


Technopigg:
Excellent work , Ill give it a proper test when back around tomorrow if Vamp hasn't by then.

Technopigg (approved):
Excellent work, ability looks great! Tested fringe cases such as does it effect Control Points etc and all looking good.  Approved

alrun3:
> Excellent work, ability looks great! Tested fringe cases such as does it effect Control Points etc and all looking good. Approved

Please squash-merge the PR at your leisure :raised_hands:

PR #4060, opened 2026-07-21
Muradin's Second Wind and Thunder Crack

ErickSebas:
Implements the two custom abilities from #4051 using the WCSharp.Buffs system.

## Second Wind
UnitTrait with two buffs: an invisible one counting the 7s out-of-combat window,
and a visible one ticking the heal (2/4/6/8% max HP per level, doubled below 40%).
Taking damage kills the regen and restarts the timer.

## Thunder Crack
Damage, AoE and slow are handled by the native Thunder Clap. The armor debuff is
applied via Devotion Aura applicators bound with BindAura, so it shows as a red -X
on the unit card.

Needed 4 separate single-level applicators (A004, A00O, A00R, A011) with fixed
-3/-5/-7/-9 rather than one multi-level aura. A single multi-level aura always
applied -3 regardless of level: BindAura calls UnitAddAbility (which always adds
at level 1) and then SetUnitAbilityLevel, and the aura appears to lock in its
bonus at the moment it's granted. GryphonOrbit avoids this because TP45 is
single-level. The code picks the right applicator by ability level.

## Fixes to existing object data
- Icon paths on TP46/TP47/TP48/TP51/TP53 pointed to files that don't exist in the
  base game or in the map's Imports (BTNHeroThunderclap, PASPowerFist,
  PASHeroThunderclap2). Swapped for BTNThunderclap and BTNRegenerate.
- TP51 and TP46 were missing the ahky field, so they inherited C and E from their
  base abilities instead of W and E.

## Shared code
Added IEffectOnAbilityLearned to MacroTools, wired to HeroTypeEvent.LearnsSpell in
UnitTypeTraitRegistry (same pattern as the existing FinishesRevive wiring). Without
it, Second Wind never started the first time it was learned, since OnDamaged bails
out at level 0 and nothing else created the initial cooldown buff. IEffectOnLearn
already exists in MacroTools.Spells but only fires for registered Spell classes, and
Second Wind is a passive UnitTrait.


Technopigg (approved):
Excellent work, very nicely done! Approved and merged

PR #4061, opened 2026-07-21
Spanish localization for quests and system text

ErickSebas:
Adds automatic Spanish localization. Detects the player's WC3 client language
and shows quest text, objectives and system messages translated, no setup
needed. 

All translations live in a single dictionary in Loc.cs (English key → Spanish
value). Fixed ~25 quests that were building their reward text with native C#
interpolation ($"...") instead of Loc.Format, which silently broke translation
since the resulting string never matched the dictionary key.

Chinese support is wired up but empty (ready for future translation).

Scope: places/zones and system text get translated. Units, buildings,
abilities, items, heroes and Control Points stay in English.

### DIFF
--- a/src/MacroTools/Commands/CommandManager.cs
+++ b/src/MacroTools/Commands/CommandManager.cs
@@ -3,6 +3,7 @@ using System.Collections.Generic;
 using System.Collections.ObjectModel;
 using System.Linq;
 using MacroTools.Extensions;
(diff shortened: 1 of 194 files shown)


Technopigg (approved):
Excellent work once again! 

Tested in Espanyol and all working as expected!

Approved

PR #4076, opened 2026-08-02
chore: bump wcsharp to 3.3.9

alrun3:
Updates to the latest version of WCSharp that adds a catch block to all `Delay` function calls


Technopigg (approved):
Thanks for doing that!

PR #4079, opened 2026-08-02
add optional per-player smart follow

grolmash:
this adds an optional smart follow system for units following a friendly hero.

the problem is that the normal moving-target follow can make large groups bunch up, take bad routes or get stuck while constantly trying to reach the hero's current position.

when smart follow is enabled, followers don't all chase the same moving point. when the hero receives a supported movement order, the system gives each follower a stable destination around the hero's destination. it tries to preserve the army's rough formation instead of sending everything into one spot.

the setting is per-player and saved between games:
  - `-smartfollow on`
  - `-smartfollow off`

native follow is the default for now. one player's setting doesn't affect anyone else.

implementation details:

  - only friendly non-hero ground units can attach
  - heroes, workers, flying units (useless: native follow works fine), boats and structures are excluded
  - followers are tracked separately for each hero
  - manual orders detach a follower from smart follow
  - move, attack-move, stop and hold-position orders are mirrored
  - unsafe or unsupported orders are left to Warcraft's native behavior
  - shift-queued hero routes are left native because later route steps can't be mirrored reliably
  - deaths, transports, removals and ownership changes clean up tracking
  - walkability checks only help choose a safe destination around the hero; Warcraft still handles the actual pathfinding
  - destinations are spread out and distant old formation positions are limited so stragglers aren't sent far away from the hero
  - follower orders are limited to 12 per game tick to avoid ordering a whole large army in one frame. i think this is particularly important, as IMO most of the non-standard "FPS" issues in legacies are due to tick saturation 

turning the setting off clears that player's tracked followers and pending smart orders without touching other players. 

i'm deliberately keeping native follow as the default until this has been tested in real games. i'd like to get experienced players using it, check whether the movement parameters need adjustments, and look for any FPS impact in an already performance-heavy map.

some ideas for making it lighter if needed: mainly grouping several follower orders into native Warcraft group commands. i'm keeping those performance experiments out of this PR so this change stays understandable and we can measure each optimization separately.

if player feedback and performance are good, i think smart follow could eventually become the default.

validation done locally: i've played the map with it (solo) and it's very nice, also tried the toggle commands and they work.

## possible performance costs and optimizations

these are possible costs, not confirmed performance problems yet. i want to measure them in real games before changing the design. 

the main things that could be expensive are:

  - each follower currently receives its own unit order. the system processes at most 12 per tick, but Warcraft still has to calculate movement for every ordered unit
  - each follower can perform up to 25 terrain checks when choosing its destination. that means up to 300 checks in one 12-unit tick, or around 5,000 across a 200-unit army in the worst case. these checks happens only when there is blocking terrain around the hero destination, to compute the best free place to go to. they can be reduced easily.
  - a pending-order object is created for every follower when the hero moves. 

  possible optimization paths:

  - use Warcraft's native group-order command for batches of followers. for example, 12 followers could receive one group order instead of 12 individual order calls
  - give each batch one shared destination, such as the hero destination or the centre of that part of the formation
  - check only the final destination instead of scanning the full line between the hero destination and follower destination
  - keep the terrain scan but increase its spacing from 32 to 64 or 96, reducing the number of checks
  - reuse pending-order data instead of creating a new Lua object for every follower
  - ignore exact duplicate hero orders that would produce the same destinations (protection against spam clicking)
  - use an adaptive order limit: keep 12 normally, but temporarily use 24 or 32 when a very large queue is falling behind, as long as this doesn't create worse single-frame pauses

the group-order idea is probably the most promising. one Warcraft group command may reduce Lua work, terrain checks and native-call overhead.


src/MacroTools/Save/SaveManager.cs
```diff
@@ -56,6 +56,7 @@ private static void SaveManager_OnSaveLoaded(PlayerSettings save, LoadResult loa
       save.ShowQuestText = true;
       save.PlayDialogue = true;
       save.ShowCaptions = true;
+      save.SmartFollowEnabled = false;
```
alrun3: `bool` already defaults to `false`

```suggestion
```

src/WarcraftLegacies.Source/Commands/Settings.cs
```diff
@@ -28,6 +28,8 @@ public override string Execute(player commandUser, params string[] parameters)
 
 Play dialogue: " + playerSettings.PlayDialogue + @"
 
-Show captions: " + playerSettings.ShowCaptions;
+Show captions: " + playerSettings.ShowCaptions + @"
+
+Smart follow: " + (playerSettings.SmartFollowEnabled ? "On" : "Off");
```
alrun3: nit: format differs from the other toggles

src/WarcraftLegacies.Source/GameLogic/AssistedFollow/AssistedFollowSystem.cs
```diff
@@ -0,0 +1,713 @@
...
+  private static readonly List<PendingBinding> _pendingBindings = new();
+  private static List<FollowerState> _pendingFollowerStates = new();
+  private static readonly HashSet<unit> _internallyOrderedUnits = new();
+
+  private static FollowOrderMode _mode = FollowOrderMode.StableDestination;
+  private static bool _isSetup;
+  private static int _activeFollowerCount;
+  private static int _pendingFollowerStateIndex;
+
+  /// <summary>The active implementation, exposed so the test harness can perform an A/B comparison.</summary>
```
alrun3: What's the test harness? The toggle behavior is superseded by the `SmartFollowEnabled` toggle. 

The static property `Mode` and its backing field `_mode` is unused (when removed from `IsPlayerEnabled`) and can be removed. Same goes for `FollowOrderMode`

src/WarcraftLegacies.Source/GameLogic/AssistedFollow/AssistedFollowSystem.cs
```diff
@@ -0,0 +1,713 @@
...
+        return;
+      }
+
+      ResetTracking();
+      _mode = value;
+    }
+  }
+
+  /// <summary>The number of units currently bound to a friendly hero.</summary>
+  public static int ActiveFollowerCount => _activeFollowerCount;
```
alrun3: The static property `ActiveFollowerCount` and its backing field `_activeFollowerCount` is unused and can be removed

src/MacroTools/Extensions/PlayerData.cs
```diff
@@ -51,6 +51,9 @@ public void UpdatePlayerSetting(string setting, bool value)
       case "ShowCaptions":
         PlayerSettings.ShowCaptions = value;
         break;
+      case "SmartFollowEnabled":
```
alrun3: nit: `nameof(PlayerSettings.SmartFollowEnabled)` would be safer, but can be followed up separately

src/WarcraftLegacies.Source/GameLogic/AssistedFollow/AssistedFollowSystem.cs
```diff
@@ -0,0 +1,713 @@
...
+    else if (_pendingFollowerStateIndex >= 512)
+    {
+      var compactedStates = new List<FollowerState>();
+      for (var i = _pendingFollowerStateIndex; i < _pendingFollowerStates.Count; i++)
+      {
+        compactedStates.Add(_pendingFollowerStates[i]);
+      }
+
+      _pendingFollowerStates = compactedStates;
+      _pendingFollowerStateIndex = 0;
```
alrun3:
```suggestion
      CompactPendingFollowerStates();
```

src/WarcraftLegacies.Source/GameLogic/AssistedFollow/AssistedFollowSystem.cs
```diff
@@ -0,0 +1,713 @@
...
+    var followerY = movementOrder.PreserveFormation ? follower.Y : leader.Y;
+    QueueFollowerOrder(state, new PendingFollowerOrder(movementOrder.FollowerOrderId,
+      leader.X, leader.Y, followerX, followerY, movementOrder.X, movementOrder.Y));
+  }
+
+  private static void GetPointOrderDestination(FollowerState state, PendingFollowerOrder pendingOrder,
+    out float destinationX, out float destinationY)
+  {
+    var worldBounds = Rectangle.WorldBounds;
+    var leaderDestinationX = System.Math.Clamp(pendingOrder.LeaderDestinationX,
```
alrun3: nit: import system

```suggestion
    var leaderDestinationX = Math.Clamp(pendingOrder.LeaderDestinationX,
```

src/WarcraftLegacies.Source/GameLogic/AssistedFollow/FollowDestinationPlanner.cs
```diff
@@ -0,0 +1,101 @@
...
+/// <summary>
+/// Produces stable follower destinations while retaining the army's current formation.
+/// </summary>
+public static class FollowDestinationPlanner
+{
+  /// <summary>
+  /// Prevents a straggler's old position from producing a destination far away from its leader.
+  /// </summary>
+  public const float MaximumFormationRadius = 768;
+
```
alrun3: `MaximumFormationRadius` has no external callers

```suggestion
  private const float MaximumFormationRadius = 768;
```

src/WarcraftLegacies.Source/GameLogic/AssistedFollow/FollowDestinationPlanner.cs
```diff
@@ -0,0 +1,101 @@
...
+  /// Projects a follower's current offset from its leader onto the leader's new destination.
+  /// </summary>
+  public static void GetDestination(float leaderX, float leaderY, float followerX, float followerY,
+    float leaderDestinationX, float leaderDestinationY, int formationSlot, out float destinationX,
+    out float destinationY)
+  {
+    var offsetX = followerX - leaderX;
+    var offsetY = followerY - leaderY;
+    var offsetDistanceSquared = offsetX * offsetX + offsetY * offsetY;
+    var maximumRadiusSquared = MaximumFormationRadius * MaximumFormationRadius;
```
alrun3: `maximumRadiusSquared` is recomputed every call from two consts and can be hoisted to `private const float MaximumRadiusSquared = MaximumFormationRadius * MaximumFormationRadius;`

```suggestion
```

src/WarcraftLegacies.Source/GameLogic/AssistedFollow/FollowDestinationPlanner.cs
```diff
@@ -0,0 +1,101 @@
...
+{
+  /// <summary>
+  /// Prevents a straggler's old position from producing a destination far away from its leader.
+  /// </summary>
+  public const float MaximumFormationRadius = 768;
+
+  private const float MinimumFormationSeparation = 48;
+  private const float BaseFallbackRadius = 96;
+  private const float FallbackRadiusPerSlot = 32;
+  private const double GoldenAngleRadians = 2.399963229728653;
```
alrun3: If you move `GoldenAngleRadians` to `float`, you can use `MathF.{Cos,Sin,Sqrt}` without a cast

```suggestion
  private const float GoldenAngleRadians = 2.399963229728653f;
```

alrun3 (approved):
LGTM cc: @vampirika, @Technopigg for approval

src/WarcraftLegacies.Source/GameLogic/AssistedFollow/AssistedFollowSystem.cs
grolmash: Uncommitted local tests (manually spawns tons of units on a map and makes them follow a hero); I'm removing all mentions

src/WarcraftLegacies.Source/GameLogic/AssistedFollow/FollowDestinationPlanner.cs
grolmash: neat

grolmash:
thx @alrun3 for the suggestions, i implemented all of them

Technopigg:
Excellent stuff!, thanks for providing a detailed summary too, super helpful!

Tested in game and I really like it. Agreed that leaving native follow as default is the best approach for now. 

@alrun3 Thanks for doing a comprehensive review. 

as discussed we will work this setting (and other local player settings) into a UI feature in the future also. 

Approved and merged

Technopigg (approved):

PR #4080, opened 2026-08-03
replace Spell Conduction buffs with damage-time checks

grolmash:
Spell Conduction used to create one buff for every Eradicator and every nearby ally. with 12 Eradicators and 200 units thats 2400 buffs running through the 32hz buff loop (77k updates/sec while _idle_)

this PR removes all those buffs. since theres only 12 Eradicators max, we just check them when a unit takes damage

tested ingame before (~10 fps drop with 12 eradicators in a mass of 200 units, infinite fps drop with 40 erads in the same amount of units)

after: unlimited eradicators are allowed


src/WarcraftLegacies.Source/Factions/Ahnqiraj/UnitTraits/SpellConduction/SpellConductionTrait.cs
```diff
@@ -25,16 +31,59 @@ public sealed class SpellConductionTrait : UnitTrait, IEffectOnCreated
...
+  public void OnCreated(unit createdUnit) => _eradicators.Add(createdUnit);
+
+  private void OnDamageTaken()
   {
-    var aura = new SpellConductionAura(createdUnit)
+    var target = @event.Unit;
+    var attackType = @event.AttackType;
+    if (_eradicators.Count == 0 ||
+        !RedirectableAttackTypes.Contains(attackType) ||
+        UnitTypeTraitRegistry.UnitHasTrait(target, typeof(SpellConductionTrait)))
```
alrun3, line 54: `UnitHasTrait`checks whether the target's unit type has a `SpellConductionTrait` registered, not whether `target` is one of _this_ instance's eradicators

```suggestion
        _eradicators.Contains(target))
```

alrun3 (approved):
LGTM cc: @vampirika, @Technopigg for approval

src/WarcraftLegacies.Source/Factions/Ahnqiraj/UnitTraits/SpellConduction/SpellConductionTrait.cs
grolmash: if you don't mind too much, i would keep this check as is. on normal situation, right now, they are both the same. but if one eradicator wasn't properly in `_eradicators` (due to some revive or who knows what), then the damage would loop. same issue if spell conduction was given to another unit than eradicator.

alrun3: Of course, no problem. Though I think there's some ambiguities we should clear up:

  - The variable name `_eradicators` implies a single 1:1 relationship, I propose we rename that to `_conductors` to convey "units currently able to redirect damage via this trait", not just "eradicators"
  - WCSharp implements `UnitTypeEvent.IsCreated` as `RegionUnitTypeEvent.Enters` over the whole map, I _think_ this also fires on resurrections, worth checking. Because if so, best case we'll see double damage (dead conductors only get pruned lazily, so a resurrected one could get added twice) and worst case the trait won't function on resurrected conductors if resurrections don't refire it

grolmash: i added some code to not duplicate eradictors and renamed the list to `_conductors` (pretty cool name btw)

Technopigg:
@grolmash Excellent work , this has been a troublesome ability from its creation. Couple fringe cases I want to test first before merging. 

Do we know how this works when encountering abilities that use the native Max Damage field? Shockwave for example or Blizzard

grolmash:
@Technopigg checked the old issue, i don't think the refactor fixes this. it still reduces the native damage event and deals the redirected part separately, so Shockwave still doesnt count the redirected damage toward its Max Damage.

Technopigg:
> @Technopigg checked the old issue, i don't think the refactor fixes this. it still reduces the native damage event and deals the redirected part separately, so Shockwave still doesnt count the redirected damage toward its Max Damage.

Fair enough. The efficiency improvements still make the refactor worth it I would say. I will dig up the old convo in Discord regarding this as there were a couple possible solutions that may be worth looking at.

Technopigg (approved):

PR #4081, opened 2026-08-04
fix: stop double-localizing GetObjectName results

alrun3:
`GetObjectName` already returns the localized name of the object. Use the value as-is instead. We lose some translated objects (and gain some), but these should have been translated using the native `wts` instead


Technopigg (approved):
Approved thanks! so removing the extra Loc.Get wrapping is better here because GetObjectName already returns the localised name? Just want to make sure I understand it correctly.

alrun3:
> [..] because GetObjectName already returns the localised name?

Exactly, see: https://github.com/lep/jassdoc/blob/46fecb9/common.j#L1775-L1793. We need to translate object data separately

alrun3:
It's worth noting that these values ​​may already be translated and therefore not match the English translation keys, so this is also a bug fix

Technopigg:
> > [..] because GetObjectName already returns the localised name?
> 
> Exactly, see: https://github.com/lep/jassdoc/blob/46fecb9/common.j#L1775-L1793. We need to translate object data separately

Gotcha makes sense thank you!

PR #4082, opened 2026-08-04
fix: lookup power by type instead of localized name

alrun3:
If the player's language changed between two resolutions of the power's name `TryGetPowerByName(Loc.Get("Cult Spies")`, would not match and the lookup would fail silently. Use `GetPowerByType<PlayerVisionPower>()` to decouple from localization

### DIFF
--- a/src/WarcraftLegacies.Source/Factions/Scourge/Quests/QuestPlague.cs
+++ b/src/WarcraftLegacies.Source/Factions/Scourge/Quests/QuestPlague.cs
@@ -6,6 +6,7 @@ using MacroTools.Localization;
 using MacroTools.Quests;
 using MacroTools.Utils;
 using WarcraftLegacies.Source.Factions.Scourge.Mechanics;
+using WarcraftLegacies.Source.Factions.Scourge.Powers;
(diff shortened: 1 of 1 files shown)


alrun3:
When I think about this, it would desync since it is matched by the display string

Technopigg:
Using `GetPowerByType` would stop any silent fails which is awesome, does this resolve the potential desync issue also do we know?

alrun3:
> Using `GetPowerByType` would stop any silent fails which is awesome, does this resolve the potential desync issue also do we know?

Yes, the _problem_ is that the name is used for localized presentation, but also for lookup. Let's say the power's title is created in language setting A and I switch to language setting B. The name will no longer match unless it is updated, so "silently fail" can lead to desynchronization, unless there's an update routine associated with language switching to keep them synchronized. I haven't checked this, so this is primarily a correctness fix that also prevents subtle desyncs

Technopigg:
ah I see, no such toutine does exist after checking. Cool sounds good Ill merge it in

Technopigg (approved):

PR #4083, opened 2026-08-04
refactor: simplify Loc.Format token substitution

alrun3:
Replace `ReplaceLiteral`'s manual token-replacement loop with `string.Replace`, and extract the local player's resolved language into `Loc.GetLanguage`


Technopigg (approved):
Solid refactor looks much easier to maintain now. Thanks for adding in the tests too !

PR #4084, opened 2026-08-04
fix: remove duplicate collision effect on muradin's stormbolt

alrun3:
I _think_ that when we implemented this there was no on-hit effect on the "dummy" ability, but when we merged the object data together we ended up with two effects. This removes the duplicate effect from the code


Technopigg:
good catch!

Technopigg (approved):

Issue #3999, opened 2026-04-05
Reimplement Dalaran Way Gate functional


YakaryBovine:
As per #3998, dynamic Way Gates are now a no-go. Dalaran Way Gate is definitely functionality we want to retain (even when they're replaced with Theramore), so we need to reimplement it without using dynamic Waygates.

Needs some thought, but current idea is to give the Way Gate a dummy sellable unit that, when bought, teleports all units in a radius around the Way Gate to the sister Way Gate.

Issue #3998, opened 2026-04-05
Temporarily disable all dynamic Way Gates except Legion's


YakaryBovine:
As far as we're aware, there is an undocumented Warcraft III bug wherein setting a Waygate's destination during a game can cause it to erroneously be instead set to the same destination as the most recently modified Waygate. As this is potentially disastrous to gameplay, we intend to temporarily disable the following features that leverage it:
- Dalaran's Way Gate
- Illidari's Slipstream to Black Temple
- Frostwolf's Portal to Nagrand

Legion doesn't function without their Slipstream, so we'll leave that one in tact. With no other sources of dynamic Waygates, it should be okay.

PR #4087, opened 2026-08-04
fix dalaran waygates

### LINKED ISSUE #3789: Dalaran Waygate only lets you go 1 way

grolmash:
brings back the 2 dalaran waygates and makes them work both ways again

a lot of the files arent new, they were removed in `137101af90398f8953ac841a84f1c3665e037a2c` when the gates got 
disabled (#3998). this restores the old unit data, farmer build option, limits, constants etc

the new version disregards the native warcraft waygate destination system which is bugged (#3789)

now the gate has a travel button. using it teleports nearby allied units to the other completed gate. it works the same in both directions

should close #3789 and #3999 (i was confused by the dummy sellable, but did do the manually teleporting all friendly units around the waygate when clicking the tp spell on gate)

tried ingame and it worked


vampirika (approved):
Seems fine to me, worth a try over trying to get native to work

PR #4088, opened 2026-08-04
fix mmd

ErickSebas:


Technopigg (approved):
Nice work , thanks for getting those stragglers hooked up. Tested in game and MMD raw data shows it now works as expected! Approved and merged

PR #4089, opened 2026-08-06
feat: consolidate cheats into composite commands to improve dx

alrun3:
I found the "cheats" hard to discover and inconsistent to use. This PR adds a composite verb-based command style plus shared chat text-layout infrastructure, so cheats are grouped under a handful of self-documenting commands and get consistent usage generation, error messages, and pagination

* Add `CompositeCommand` that dispatches commands to named verbs (e.g. `-faction set`, `-quest list`), automatically generates its usage/verb-list text, and paginates the output of any command that would overflow the chat window so that individual commands don't have to handle either
* Add `CommandTargets` to resolve chat-typed faction/quest/power/team parameters either by their list position or class name
* Add `MacroTools.Chat` (`GameText`, `Pager`, `ColumnFormatter`) to approximate WC3's chat font, paginate long output, and build column-aligned output
* Add `CheatHelp` to list registered cheats
* Consolidate `CheatTeam`, `CheatControl`, `CheatQuestProgress`, `CheatRemovePower`, and `CheatShowQuestNames` into three composite commands: `CheatFaction`, `CheatQuest`, `CheatPower`
* Fix `CommandManager` to return no parameters instead of throwing when the raw input is shorter than the command prefix
* Deprecate `GetQuestByTitle` and `TryGetPowerByName` as these lookup localized properties


Technopigg (approved):
Nice work actually much easier, I especially like the Help command :D 

Approved!

PR #4090, opened 2026-08-06
fix: avoid mutating collection during iteration

alrun3:
`FontOfPower` and `Immortality` calls `RemoveObjective` while foreach-iterating the same list, throwing `InvalidOperationException`:

```
Failed to execute command: System.InvalidOperationException: Collection was modified; enumeration operation may not execute.
war3map.lua:20256 <- 67 <- 5753 <- 5787 <- 44909 <- 36499 <- 28334 <- 30706 <- 30265 <- 89 <- 30245
```

The call to `_objectives.Clear()` already empties the list, so teardown only needs to unsubscribe the event handler


Technopigg:
Nice catch , surprised this wasn't causing crashes.

Technopigg (approved):

PR #4092, opened 2026-08-08
fix: use localized fallback in quest flavour text

alrun3:


Technopigg (approved):
Nice, thanks. I really ought to of let you review the localisation PR rather than my skimming over it and peforming basic in game tests!

PR #4093, opened 2026-08-08
refactor: replace .First() with indexer access on collections

alrun3:


Technopigg:
@alrun3 are you able to resolve the merge conflicts on this one. Also a bit out of my depth here, is it just more efficient than the current LINQ method `.First`?

alrun3:
Yes to both, we can also drop it if you don't think it adds value

Technopigg:
I trust you with anything like this, I had a little read and from what I could understand `.First` is the generic all use way to query any sort of collection but it is that same flexibility of use that means the overhead from using it is a bit more than what you have suggested! lmk if that is the case, will merge anyway, I am just interested :)

alrun3:
> [..] from what I could understand `.First` is the generic all use way to query any sort of collection but it is that same flexibility of use that means the overhead from using it is a bit more than what you have suggested! lmk if that is the case, will merge anyway, I am just interested :)

Spot on! I'll fix the conflict tomorrow

alrun3:
Rebased on `dev`

Technopigg (approved):

PR #4094, opened 2026-08-09
refactor: remove redundant player settings defaults

alrun3:
Deserialization always builds a fresh instance via `Activator.CreateInstance` and only overwrites properties present in the save data, so per-field defaults already apply on load failure and on missing/legacy fields


Technopigg (approved):
Nice clean up!

PR #4095, opened 2026-08-10
fix: correct Thunder Crack's target filter, debuff duration, and tooltip

alrun3:
* Filters magic-immune enemies from the armor debuff
* Syncs Thunder Crack's armor debuff duration with its native ability data
* Updates the tooltip to show the per-level/hero duration, and drops redundant newlines that were doubling up line breaks between levels

### DIFF
--- a/mapdata/WarcraftLegacies/AbilityData/TP51.json
+++ b/mapdata/WarcraftLegacies/AbilityData/TP51.json
@@ -78,7 +78,7 @@
     {
       "Id": "arut",
       "Type": 3,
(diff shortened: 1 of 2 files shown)


Technopigg (approved):
Nice tidyup and magic immune units did not even occur to me for the debuff side of things!

alrun3:
> Nice tidyup and magic immune units did not even occur to me for the debuff side of things!

I don't think our custom spells respect magic immunity. I wawnted to add the check to the general `IsTargetEnemyAliveAndGroundUnits` filter, but that would affect the other spells that use it. Maybe I should log an issue for investigation as a follow-up? I think there were 2-3 cases in total

Technopigg:
Sounds like a good change for me , only three cases to change is nice Sent from my iPhoneOn 11 Aug 2026, at 20:17, alrun3 ***@***.***> wrote:﻿alrun3 left a comment (AzerothWarsLR/WarcraftLegacies#4095)

Nice tidyup and magic immune units did not even occur to me for the debuff side of things!

I don't think our custom spells respect magic immunity. I wawnted to add the check to the general IsTargetEnemyAliveAndGroundUnits filter, but that would affect the other spells that use it. Maybe I should log an issue for investigation as a follow-up? I think there were 2-3 cases in total

PR #4096, opened 2026-08-10
feat: add LocalizedText for deferred, reactive localization

alrun3:
Names, titles, and descriptions are currently localized eagerly against the system's locale, so they can't respond to the in-game language command. This PR adds classes to store the template and substitution args instead of a resolved string, so text can be re-resolved on demand whenever the player's selected language changes

* Splits the hardcoded Spanish dictionary out of `Loc` into a `SpanishTranslation : ITranslation` so tests can swap in fake translations and reset shared static state
* Adds `LocalizedText` and `LocalizedTextArg` for storing template and placeholders to enable reactive, recursive localization

If this lands, I'll follow up with wiring into properties of callers (Faction, Quest, Power etc.)


Technopigg (approved):
Nice , I agree this is much cleaner. Please go ahead with the follow up changes

PR #4099, opened 2026-08-17
perf: move unit-target dummy casters directly

### LINKED ISSUE #4098: MassAnySpell performance investigation
legacies has fps issues, but those never get as worse as when people start spamming spells during fights

### frost armor and general explanations

many (most?) mass spells use the MassAnySpell function, here's it how it roughly works:

when casting a mass spell like **frost armor** on 200 units, for each unit:
  1. get target position  ~0.4ms
  2. assign dummy owner ~0.4ms
  3. move dummy to target (**~22.8ms** -> now **~0.4ms**)
  4. give dummy spell + assign level **~4.4ms**
  5. cast spell **~24ms**
  6. remove spell ~1.2ms
  7. enumeration/other stuff ~0.8ms
on the right are the cumulated times for 200 units

total: around **54ms** per cast for frost armor on 200 units

so basically 90% of the entire time was just moving the dummy and ordering it to cast

if 12 units did this at once against 200 targets it is around 650ms of processing, so if wc3 stops to do the computation, as it surely does, this will be a 0.65 second global lag spike

i will try to bring down the time per cast to a minimum, starting from 54ms. here are some ideas to try out

- [x] moving dummy to target currently takes a huge time bc it does position checks (SetUnitPosition) that are not even required as the dummy is flying. replacing this with (SetUnitX + SetUnitY) + tested out: it gets rids almost completely of the moving times, makes us go from **54ms -> 30ms** (merged in #4099)
- [ ] **PR inc** giving the dummy the spell + level only once per mass cast works: all 200 orders still completed successfully, bringing the average cast time **30ms -> ~26ms**
- [ ] investigate whether some of the 200 individual cast orders can be removed entirely. these orders alone take around 24ms, so getting significantly below 25ms is impossible while still issuing one native order per target

current result is around 26ms/cast, meaning the theoretical 12-caster spike went from roughly 650ms -> 320ms

discarded idea: replace every MassAnySpell with a native wc3 mass spell. 23 different spells currently use this system, with completely different effects, so... i prefer optimizing the general case, every ms shaved off will get a multiplier

## 6 dmg + stun massanyspells

War Stomp, Doom Lord War Stomp, Seismic Shard, Scattershot, Enchanted Bolts, Mass Ice Lance

these are 6 abilities  that are damage + stun + animation. they can be very slow, for instance war stomp does both thunder clap + storm bolt each unit. 

- [ ] we catch all 6 by using fan of knives for damage (range + dmg are parametrized) + code for the stun and stun effect.

#### war stomp
warstomps take double the usual time bc it applies damage + stun individually

1. resolve dummy cast position        ~1.0ms
2. get + assign dummy owner           ~2.1ms
3. move dummy                         ~1.0ms
4. give dummy spell + assign level    ~4.7ms
5. face dummy toward target           ~1.0ms
6. order dummy to cast stun          ~26.4ms
7. remove spell                       ~1.0ms
8. damage/enumeration/effect/other    ~22.2ms

Total: 59.4 ms average per cast (nb: before the setposition changes, this would have been  104ms...)

## other massanyspells

8 
Fel Horde Thunder Clap, Gilneas Thunder Clap, Consecration, Void Toxin, Storm Surge, Crippling Shout, Mass Enrage, Mass Unholy Frenzy

9
Blinding Starlight, both Mass Banishes, Discordant Cadenza, Mass Ensnare, Crusader’s Shout, Mass Anti-Magic Shell, Elune’s Gaze, Mass Frost Armor

## other mass spell candidates (not using massanyspell)

stormwind's electric strike

grolmash:
SetPosition was being used to move the flying, zero-collision global dummy before every unit-target cast. this performs general positioning/pathing checks that the dummy doesn't need

using direct X/Y positioning reduced the movement cost for 200 targets from ~23ms to ~0.4ms, bringing a MassAny cast on 200 units from ~54ms to ~30ms

tested ingame across repeated 200-target Mass Frost Armor casts, with all 200 targets affected every time

### DIFF
--- a/src/MacroTools/DummyCasters/GlobalDummyCaster.cs
+++ b/src/MacroTools/DummyCasters/GlobalDummyCaster.cs
@@ -18,7 +18,8 @@ public sealed class GlobalDummyCaster
     var originPoint = originType == DummyCastOriginType.Caster ? caster.GetPosition() : target.GetPosition();
     var owningPlayer = caster.Owner;
     _unit.SetOwner(owningPlayer);
-    _unit.SetPosition(originPoint.X, originPoint.Y);
+    _unit.X = originPoint.X;
+    _unit.Y = originPoint.Y;
     _unit.AddAbility(abilId);
     _unit.SetAbilityLevel(abilId, level);
 


Technopigg (dismissed):
Great catch, nice work! I wonder why Yak set it up that way. I checked with Lord Damage and Orden and they said the same, just use X/Y (unless for whatever reason you needed Z ever too)

Technopigg (approved):

PR #4100, opened 2026-08-17
Vampirika/balance na focus

vampirika:
https://docs.google.com/document/d/1Bjcf-X1fBBYzFl6pRGfQRi0JAFKx9L-zfEZRN-xiEHY/edit?tab=t.0#heading=h.75wbz96i9ot0


src/WarcraftLegacies.Source/Factions/Lordaeron/LordaeronTraits.cs
```diff
@@ -13,5 +15,24 @@ public static void Setup()
...
+    var scourgeBaneConditions = new List<DamageCondition>
+    {
+        new() {
+          Damage = 100,
+          Condition = target => target.Race == RACE_UNDEAD && target.IsUnitType(unittype.Summoned),
+          Effect = "Abilities\\Spells\\Human\\HolyBolt\\HolyBoltSpecialArt.mdl"
+        },
+        new() {
+          Damage = 33,
+          Condition = target => target.Race == RACE_UNDEAD,
```
Technopigg: target.Race == Race_UNDEAD refers to any unit flagged as Undead in the object editor , annoyingly this is'nt actually exclusively used for Undead units at the moment (this procs on the Fel orcs in Alterac for example)

src/WarcraftLegacies.Source/Factions/Lordaeron/LordaeronTraits.cs
```diff
@@ -13,5 +15,24 @@ public static void Setup()
       DummyOrderId = ORDER_RESURRECTION,
       ProcChance = 0.2f
     }, UNIT_H01J_THE_ASHBRINGER_LORDAERON);
+
+    var scourgeBaneConditions = new List<DamageCondition>
+    {
+        new() {
+          Damage = 100,
+          Condition = target => target.Race == RACE_UNDEAD && target.IsUnitType(unittype.Summoned),
+          Effect = "Abilities\\Spells\\Human\\HolyBolt\\HolyBoltSpecialArt.mdl"
```
Technopigg, line 25: Looks awesome but its happening so often, can we make this passive have a % chance to proc rather than proccing on each and every hit? it is also very very strong currently

Technopigg (changes requested):

src/WarcraftLegacies.Source/Factions/Lordaeron/LordaeronTraits.cs
vampirika: Now uses "UnitClassifications.Undead" which is only applied to relevant creeps as most creeps have Creep race, and have added undead classification to Lord Barov.

src/WarcraftLegacies.Source/Factions/Lordaeron/LordaeronTraits.cs
vampirika: Added proc chance (currently 40%) and reduced damage slightly for regular undead, as well as the reduced attack speed compared to live. This results them in being worse against plain units, almost equal against undead but much better against summoned undead e.g. skeletons and animate dead

Technopigg (approved):

PR #4120, opened 2026-08-24
Tauren orc foundation

Technopigg:
Sets up Tauren Tribes and Orcish Horde as factions with supporting classes for the skeleton.


vampirika (approved):
Good place to start

PR #4134, opened 2026-09-05
perf: extra position improv + less object created

grolmash:
Small fix to dummy cast, similar to previous #4099 

there were other casts using SetPosition instead of direct setting

i also remove the amount of object creation, mainly creating points when floats should simply be passed around.

1. in faceposition, a point should not be created and destructed again

2. same for originPoint: instead of creating an object (heap allocation = bad), directly set the x/y

### DIFF
--- a/src/MacroTools/DummyCasters/AbilitySpecificDummyCaster.cs
+++ b/src/MacroTools/DummyCasters/AbilitySpecificDummyCaster.cs
@@ -32,7 +32,7 @@ public sealed class AbilitySpecificDummyCaster
 
     if (originType == DummyCastOriginType.Caster)
     {
-      _unit.FacePosition(target.GetPosition());
+      _unit.FacePosition(target.X, target.Y);
     }
 
     _unit.IssueOrder(_abilityOrderId, target);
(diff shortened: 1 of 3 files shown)


src/MacroTools/DummyCasters/GlobalDummyCaster.cs
```diff
@@ -15,17 +15,24 @@ public sealed class GlobalDummyCaster
   /// </summary>
   public void CastUnit(unit caster, int abilId, int orderId, int level, unit target, DummyCastOriginType originType)
   {
-    var originPoint = originType == DummyCastOriginType.Caster ? caster.GetPosition() : target.GetPosition();
+    if originType == DummyCastOriginType.Caster
```
alrun3, line 18: This isn't valid syntax and needs to be fixed

src/MacroTools/DummyCasters/AbilitySpecificDummyCaster.cs
```diff
@@ -56,7 +56,8 @@ public void CastNoTargetOnUnit(unit caster, int level, unit target)
   {
     var whichPlayer = caster.Owner;
     _unit.SetOwner(whichPlayer);
-    _unit.SetPosition(target.X, target.Y);
+    _unit.X = target.X;
+    _unit.Y = target.Y;
```
alrun3, line 60: As per the members' jassdoc, the difference between `SetPosition` and the `X` and `Y` properties is pathing and canceling orders. We don't want pathing, but can not canceling orders lead to unintended side effects?

alrun3: If the answer is no, is there a reason not to apply the changes to the rest of the methods in `AbilitySpecificDummyCaster`?

PR #4147, opened 2026-09-19
Feat/locale overlay build

nsYW:
### Summary

Adds first-class support for building a map in a specific language, without forking or duplicating `mapdata`.

Today `mapdata` is single-language, so a community translation has only one option: overwrite the English text in place. That makes the translated text the *only* text the published map can contain, and it makes a translation PR unreviewable as a translation — it looks like a rewrite of the map's content.

This PR introduces a **sparse locale overlay** instead:

```
mapdata/WarcraftLegacies/UnitData/E002.json        <- base, stays English
mapdata/WarcraftLegacies/UnitData.zhCN/E002.json   <- overlay, only the files a translation changes
mapdata/WarcraftLegacies/zhCN/buildtext.json       <- standalone root files
```

- `--locale <name>`: for every data directory, `<directory>.<locale>/` takes precedence over the base directory, per file; anything the overlay does not contain falls back to the base file. Standalone root files are overridden from `<root>/<locale>/`.
- **`BuildText`** reads `<root>/<locale>/buildtext.json`. Some player-facing strings are composed during a build rather than stored in the object database — the build bar's `Build {name}`, a unit's description, a unit's limit line — so they are looked up at build time from that file. A key stored under a spelling the caller does not ask for reaches the player in English, which is why the lookup is checked: a unit id is stored byte-reversed, and the tooltip migration built its key without inverting it, so every `unit description:` key missed.

### Verification

- `dotnet test test/Warcraft.Cartographer.Tests` — 5 tests, all passing (how a locale overlay's paths are resolved)
- `dotnet build -c Release src/WarcraftLegacies.CLI` — 0 errors
- with no `--locale` the base map data is used unchanged, so existing builds are unaffected: the English build of this map is what it was before, apart from the fields the tooltip migrations compose
- the merged map loads on a zhCN client with a Chinese interface, object data and map name, and on an enUS client with the English ones


nsYW:
图片
图片
图片
图片

Technopigg (dismissed):

Technopigg (approved):

PR #4148, opened 2026-09-19
Feat/Add Simplified Chinese map data overlay

nsYW:
Adds mapdata/WarcraftLegacies/*.zhCN/, a sparse overlay carrying the Simplified Chinese unit, ability, item, upgrade, buff and destructable names and tooltips - 14,184 unique strings across 33,603 field values in 3,977 files - together with the locale's root files (buildtext.json, Info.json, war3mapSkin.txt, war3mapMisc.txt).

The base mapdata is left untouched, so the English build is unchanged; build with --locale zhCN to get the Chinese map


Technopigg:
Hey big thanks for going to this effort!. 

We also have the foundations in place to localise the language of custom quest, item, intro text , hints etc. Currently its set up for Spanish and English only but having it in CN would be great!  Would that be something you might be interested in working on ? 

I'll review this and get it merged soon.

Technopigg:
I had not checked #4147 it appears you already have that covered

PR #4149, opened 2026-09-20
Tyrande lunar sanctuary

Technopigg:
New custom themed spell for Sentinels - Tyrande. AOE heal and slow that reveals invis enemy units "in the moonlight"- Replaces Crown Heal/"Words of Protection"


Technopigg:
@vampirika Could test this for balance and fit

vampirika (approved):
Pretty much all looks good. Only thing is the initial explosion visual is a bit in your face. Balance can be tuned later if an issue

Technopigg:
Cool and I'll have a check , I can't recall adding an explosion SFX so will have to check what's causing that

PR #4150, opened 2026-09-20
Emerald dream

Technopigg:
New custom themed ability for Malfurian; Emerald Dream is a type of rescue spell.


vampirika (changes requested):
I like it. Is there a way to add a sound effect, maybe another visual along with the existing just so its obvious stuff is happening to the unit.

Also, malfurion doesn't seem to be trainable from the altar - not sure if just this branch or general but it's not included in the trainable units.

Technopigg:
> I like it. Is there a way to add a sound effect, maybe another visual along with the existing just so its obvious stuff is happening to the unit.
> 
> Also, malfurion doesn't seem to be trainable from the altar - not sure if just this branch or general but it's not included in the trainable units.

Yeah can try and add a cool sound effect and/or visual. I made it not work on Cen for balance , let me know if you think thats correct or not.

RE Malf - it's because you get him via the first quest not trained at an altar.

vampirika:
> > I like it. Is there a way to add a sound effect, maybe another visual along with the existing just so its obvious stuff is happening to the unit.
> > Also, malfurion doesn't seem to be trainable from the altar - not sure if just this branch or general but it's not included in the trainable units.
> 
> Yeah can try and add a cool sound effect and/or visual. I made it not work on Cen for balance , let me know if you think thats correct or not.
> 
> RE Malf - it's because you get him via the first quest not trained at an altar.

Yeah not working on magic immune I think is fine. Has that always been the case for malfurion? ;o

Technopigg:
Yup , you gotta get the horn of cenarius and take it to the summoning pedestal Sent from my iPhoneOn 22 Sep 2026, at 20:25, B.Towner ***@***.***> wrote:﻿vampirika left a comment (AzerothWarsLR/WarcraftLegacies#4150)


I like it. Is there a way to add a sound effect, maybe another visual along with the existing just so its obvious stuff is happening to the unit.
Also, malfurion doesn't seem to be trainable from the altar - not sure if just this branch or general but it's not included in the trainable units.

Yeah can try and add a cool sound effect and/or visual. I made it not work on Cen for balance , let me know if you think thats correct or not.
RE Malf - it's because you get him via the first quest not trained at an altar.

Yeah not working on magic immune I think is fine. Has that always been the case for malfurion? ;o

PR #4151, opened 2026-09-20
Watchers Focus and Glaive Trap for Naisha 

Technopigg:
Both custom spells , one mine style ult and one aoe crit buff with a new overhead buff indicator


vampirika (changes requested):
Watcher's focus crit damage is 2/3/4/4 but tooltip states "double" damage. Also I'd consider lowering the manacost/cooldown/duration so naisha can create glaives more often from her passive, but not necessary.

Ult is cool :)

Technopigg:
> Watcher's focus crit damage is 2/3/4/4 but tooltip states "double" damage. Also I'd consider lowering the manacost/cooldown/duration so naisha can create glaives more often from her passive, but not necessary.
> 
> Ult is cool :)

Yeah sounds good to me, ill make those changes and sort the clarity out.

PR #4152, opened 2026-09-21
Anasterian spell - Sunfire Barrage

Technopigg:
A custom ultimate spell for Anasterian. This is a play on his existing defensive orbs ability. An aoe firebolt spell that adds a projectile for each additional defensive orb Anasterian has floating around him at that time "spending" the orbs when he casts it.


vampirika (approved):
Very cool, ability and concept. I would do quite a bit to balance this though, as the ultimate is "spending" the orbs which are otherwise doing a lot of damage. I think we can either:

- Have this ult have a much shorter cooldown, so that you can repeatedly throw the orbs.
- Massively tune up the aoe and damage of throwing the orbs. They didn't seem to exactly hit due to the projectile calculations and the damage seems almost less than just have the passive maxxed out. In short, it doesn't seem worth to spend the orbs (when it should).

Technopigg:
> Very cool, ability and concept. I would do quite a bit to balance this though, as the ultimate is "spending" the orbs which are otherwise doing a lot of damage. I think we can either:
> 
> * Have this ult have a much shorter cooldown, so that you can repeatedly throw the orbs.
> * Massively tune up the aoe and damage of throwing the orbs. They didn't seem to exactly hit due to the projectile calculations and the damage seems almost less than just have the passive maxxed out. In short, it doesn't seem worth to spend the orbs (when it should).

Makes sense, Ill try and clean up the projectiles abit , I had the orbs landing around the target point before intentionally but can have them all land on the same spot. For damage I was worried it was too strong already but works for me , ill buff it up a bit!

vampirika:
> > Very cool, ability and concept. I would do quite a bit to balance this though, as the ultimate is "spending" the orbs which are otherwise doing a lot of damage. I think we can either:
> > 
> > * Have this ult have a much shorter cooldown, so that you can repeatedly throw the orbs.
> > * Massively tune up the aoe and damage of throwing the orbs. They didn't seem to exactly hit due to the projectile calculations and the damage seems almost less than just have the passive maxxed out. In short, it doesn't seem worth to spend the orbs (when it should).
> 
> Makes sense, Ill try and clean up the projectiles abit , I had the orbs landing around the target point before intentionally but can have them all land on the same spot. For damage I was worried it was too strong already but works for me , ill buff it up a bit!

I think having them land around is fine, it does give it that magic-randomness factor which is cool, and will mostly be used on armies anyway. Just should feel worth it to fling the orbs in general

PR #4154, opened 2026-09-21
Rexxar - Toolkit refresh

Technopigg:
Q - Wild Throw (a version of Muradins Stormbolt for Rexxar)
W - Bestial Wrath - Attack damage and attack speed increase for Rexxar and any of his summoned pets.
E - Hardened - Unchanged 
R - Stampede - Unchanged


vampirika (changes requested):
Beastial Wrath doesn't apply to hawk. (bug)
Might be worth giving it some defense thing too, like some bonus armor or even the new magic resist? Also need to balance numbers imo (can do later date).

Axe is exactly the same as muradin hammer atm with a stun. Is that intended how we want? Feels like more damage and a slow would fit the axe better, but also fine with it being the same.

Technopigg:
Will fix the aoe not applying the hawk and re balance - up or down? I did think it was quite a lot of damage at higher levels . Just let me know and I’ll change it up. And I avoided Armor as he already has hardened but can add it if you want ? RE the axe, yeah it’s like for like atm which prob isn’t great in terms of spell philosophy- I’ll make it do a DOT maim/ slow perhaps instead of a stun ?  Sent from my iPhoneOn 22 Sep 2026, at 21:45, B.Towner ***@***.***> wrote:﻿
@vampirika requested changes on this pull request.

Beastial Wrath doesn't apply to hawk. (bug)
Might be worth giving it some defense thing too, like some bonus armor or even the new magic resist? Also need to balance numbers imo (can do later date).
Axe is exactly the same as muradin hammer atm with a stun. Is that intended how we want? Feels like more damage and a slow would fit the axe better, but also fine with it being the same.

PR #4155, opened 2026-09-21
Rivendare corruption mine

Technopigg:
**Curruption Mine**

New ultimate for Baron Rivendare, replacing his older lacklustre ultimate that could be easily dispelled.


vampirika (changes requested):
Just need to adjust primer time as discussed, otherwise all good

PR #4156, opened 2026-09-21
Antonidas mana bomb

Technopigg:
**Mana Bomb**
Ultimate ability for Antonidas replacing the now redundant Dalaran Sheild.

A large ball of energy grows in size overhead of the caster dealing more damage, up until its max, the longer it is channelled for. If the spell is interrupted during channelling the ball will launch at its current size dealing reduced damaged.


Technopigg:
Maybe need a new name for this one as it doesnt do anything to enemy mana at all

PR #4157, opened 2026-09-22
Feat/zh overlay on dev

nsYW:
original #4148


mapdata/WarcraftLegacies/AbilityData/Augh.json
nsYW: One file outside the .zhCN folders: AbilityData/Augh.json, a base record for a Blizzard rawcode that the overlay localises. Without it the overlay entry has no base object and is dropped at build time.

Technopigg (approved):

PR #4159, opened 2026-09-22
Orcish horde quests

Technopigg:
New quests:
- The Crossroads
- The Darkspear Trolls (Sen'jin / Vol'jin)
- Slay Cenarius (Grom and Grunt upgrades)
- Destroy the Elf Cities
- Thrall Maelstrom
- Warsong Hold
- Free Ner'zhul

Other changes:
- Quests unlock in order after Orgrimmar is built
- Orgrimmar quest now needs an altar, war mill and war camp
- Grom and the Siege Workshop need Orgrimmar built first
- Hero order at the altar tidied up
- Longer quest timers
- Horde Pier can research R06G
- Removed hostile units on the Echo Isles
- Removed the unused Build Orgrimmar research (


Technopigg:
@vampirika this should be ready to the point where you can start as Orcish Horde and play through all there quest in normal game order and all *should* work - so the review should also cover timings and heroes levels based on what time they end up with for balance.

PR #4162, opened 2026-09-23
Magatha kit

Technopigg:
New spells:                                                                                                                                                                                                                                                                                                       
  - Spirit Mend: chain heal that jumps to injured allies and gives them an Earthen Shield                                                                                                                                                                                                        
  - Grimtotem Ward: ward that heals nearby allies and slows nearby enemies                                                                                                                                                                                                                         
  - Crone's Curse: slows an enemy's attack speed, heals nearby allies if it dies while cursed                                                                                                                                                                                                                       
  - Wrath of the Earth Mother: channel that roots and damages enemies around her while healing allies with tranquillity.


Technopigg:
One thing I realised after this one was she has 4 active abilities, which is a bit much and kind of out of sync with the normal format. Maybe okay because she is a late game unlock, not sure really.

